PR #21, opened 2025-01-05
(feat) - booking algorithm v2

teetangh:


app/api/events/classes/route.ts
```diff
@@ -5,37 +5,23 @@ export async function GET(request: Request) {
...
     const consulteeProfileId = searchParams.get("consulteeProfileId");
-    const consultantId = searchParams.get("consultantId");
+    const consultantProfileId = searchParams.get("consultantProfileId");
 
     let classes;
 
     if (consulteeProfileId) {
       classes = await prisma.class.findMany({
         where: {
           OR: [
```
teetangh: where are the classes where the appointment is there. this is only the ones with the wait-list

app/api/events/webinars/route.ts
teetangh, line 1: the webinars with appointment is there here

app/api/events/classes/route.ts
teetangh: added

PR #26, opened 2025-03-13
(fix) booking algorithm

teetangh:
- calendar slot timings for 4 events

### DIFF
--- a/app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx
+++ b/app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx
@@ -1,19 +1,8 @@
-"use client";
 
-import { use } from "react";
 import { RequestSlotAllocationTab } from "./RequestSlotAllocationTab";
 
-export default function RequestsPage({
-  params,
-}: Readonly<{
-  params: Promise<{ consultantId: string }>;
-}>) {
-  const resolvedParams = use(params);
-  const consultantId = resolvedParams.consultantId;
+export default function RequestsPage() {
 
-  const handleUpdate = () => {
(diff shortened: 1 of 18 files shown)


app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx
Comment (teetangh) [inline app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx:1]: why are we not requiring params here ?

Comment (teetangh) [inline app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx:1]: why are we not requiring params here ?

app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx
```diff
@@ -1,19 +1,8 @@
...
 import { RequestSlotAllocationTab } from "./RequestSlotAllocationTab";
 
-export default function RequestsPage({
-  params,
-}: Readonly<{
-  params: Promise<{ consultantId: string }>;
-}>) {
-  const resolvedParams = use(params);
-  const consultantId = resolvedParams.consultantId;
+export default function RequestsPage() {
```
Comment (teetangh) [inline app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx:4]: why are we not requiring params here ?

Comment (Divyam-Kumar-Pandey) [reply app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx:4]: I changed this while I was fixing the re-rendering issue.
This part was unnecessary, and I thought this might be causing some issue, so I removed it.

Comment (Divyam-Kumar-Pandey) [reply app/dashboard/consultant/[consultantId]/(features)/requests/page.tsx:4]: But I don't think this is causing some issue, so we can revert this if you want.

teetangh:
outdated stale branch

PR #200, opened 2025-10-09
feat(middleware): implement CORS support in middleware

teetangh:
- Added a new function `addCorsHeaders` to handle CORS headers in responses.
- Updated middleware to manage OPTIONS preflight requests and include CORS headers for public, private, and protected API routes.
- Enhanced response handling to ensure proper CORS configuration based on the request origin.


teetangh:
Not necessary, I haven't faced any course issue as of now.

PR #201, opened 2025-10-10
(design) updated landing page with dark theme

teetangh:


teetangh:
A lot of work has gone into this PR, but unfortunately, we will have to close it since there has been a better UI ex that has been implemented.

PR #211, opened 2025-10-29
fix(allocation): Correct recurring event auto-allocation logic

shubham79a:
**Issues:**

-   The auto-allocation algorithm for recurring event (subscriptions) failed across different plan types (basic, extended, comprehensive).

-   **Date Filtering:** Available slots were not consistently filtered based on the subscription's precise `startDate` and `endDate` (including time), leading to attempts to select slots outside the valid period.

-   **Selection Logic:** The previous weekly loop incorrectly selected individual slots using spacing logic (`selectSlotsWithSpacing`) instead of finding consecutive blocks required for sessions. It also failed to correctly enforce the `callsPerWeek` limit at the session level and lacked the "one session per day" constraint.

-   **Loop Control:** The termination logic relied on an inaccurate `totalWeeks` calculation and tracked individual slots instead of completed sessions.

-   **Output Format:** The function previously selected 1-hour slots but returned them without splitting, causing downstream validation errors as the system requires 30-minute slot representations for saving.

-   **Missing Data:** The core `allocateRecurringSlots` function lacked access to the `options` object (containing `sessionDurationInHours`, `startDate`, `endDate`), preventing dynamic calculations.

**Fixes:**

1.  **Passed `options`:** Modified `autoAllocate` to pass the `options` object down to `allocateRecurringSlots`. Updated the `allocateRecurringSlots` signature to accept `options` and use its properties (`callsPerWeek`, `sessionDurationInHours`, `startDate`, `endDate`).

2.  **Date/Time Filtering:** Implemented strict filtering within `allocateRecurringSlots` to ensure only 1-hour slots fully contained within the `options.startDate` and `options.endDate` (inclusive of time) are considered.

3.  **Refactored Selection Strategy ("Split at End"):**

    * Replaced the old weekly loop with a "Gather, Sort, Select" approach operating on the filtered 1-hour slots.

    * Calculates `totalSessionsRequired` dynamically based on `totalSlots` and `sessionDurationInHours`.

    * Gathers all potential 1-hour slots (representing sessions) across all valid weeks.

    * Scores and sorts these potential sessions globally based on preferences.

    * Selects the best-scoring sessions greedily, strictly enforcing `totalSessionsRequired`, `callsPerWeek`, and the "one session per day" limit using a `selectedDates` Set.

4.  **Final Splitting:** After selecting the required number of 1-hour slots, added a final step using `flatMap` to convert each selected 1-hour slot into two consecutive 30-minute slots before returning the result.

5.  **Partial Allocation Handling:** Added logging to indicate if the full `totalSessionsRequired` could not be met due to constraints. The `autoAllocate` function still returns an error if `selectedSlots.length !== requiredSlots`, clearly handling partial allocations as failures.

**Probable Outcome:**

Auto-allocation for subscription and class events should now function correctly for basic, extended, and comprehensive plans. It will respect the subscription's date range, the configured `callsPerWeek`, the "one session per day" limit, select based on preferences where possible, and return the allocated times in the required 30-minute slot format, preventing validation errors during saving. Partial allocations where not enough valid slots exist will be reported as errors.

### DIFF
--- a/app/dashboard/consultant/[consultantId]/(features)/shared/utils/allocationAlgorithms.ts
+++ b/app/dashboard/consultant/[consultantId]/(features)/shared/utils/allocationAlgorithms.ts
@@ -261,6 +261,7 @@ export class AllocationAlgorithms {
             options.callsPerWeek || 1,
             options.durationInMonths || 1,
             preferences,
+            options
           );
           strategy = "optimal-distribution";
           break;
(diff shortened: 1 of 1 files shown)


shubham79a:
Hi team,

This PR implements the necessary fixes for the autoAllocate logic for recurring events (Subscriptions). I've tested this against our various plans (Basic, Extended, Comprehensive), and I can confirm the new logic is working correctly.

Summary of Fixes:

Correct Date Range Filtering: The allocateRecurringSlots function now correctly receives the subscription's startDate and endDate from the options object and filters the initial 1-hour availableSlots list. This ensures only slots fully within the subscription's valid timeframe are ever considered.

"Split at End" Strategy: As we discussed, the logic now performs all selection (grouping, scoring, filtering) on the 1-hour slots that represent one "session". Only after the correct number of 1-hour sessions have been selected does it split them into the 30-minute format required for the database, resolving the downstream validation errors.

Constraint Enforcement: The selection logic has been completely refactored to correctly enforce all business rules:

Total Sessions: It dynamically calculates the total number of sessions required (e.g., 8 sessions for 16 slots) and stops correctly when this number is met.

Weekly Limit (callsPerWeek): The new "Gather, Sort, Select" logic correctly tracks and enforces the maximum number of sessions allowed per calendar week (e.g., callsPerWeek: 2).

Daily Limit: A check has been added to ensure only one session (one 1-hour block) can be selected per calendar day, fulfilling this key requirement.

Validation & Outcome:

I've run tests with multiple scenarios, including the "Comprehensive" plan (e.g., callsPerWeek: 2 or 3) and the "Basic" plan (callsPerWeek: 1).

The algorithm successfully selects the correct number of 1-hour slots based on the plan's rules.

All restrictions (total, weekly, and daily) are being properly validated and respected.

The final returned array consists of the correct number of 30-minute slots (e.g., 12 selected 1-hour slots are correctly returned as 24 30-minute slots).

This resolves the "Not enough slots" errors and the 1-hour slot validation failure.

This should fix the auto-allocation flow for all subscription types.

PR #240, opened 2025-12-07
Pr 238

miracneroid:


teetangh:
@miracneroid Thanks for the contributions.
1. Would you be great if you made a branch of the original PR. That way i could see the diff of the Pr more easily
2. We will create a dedicated prisma fix for prisma 7 in a future PR.

PR #404, opened 2026-01-31
fix(booking): critical bug fixes, dedup, and 334 tests for slot allocation system

teetangh:
## Summary

Deep audit of the booking/slot allocation system uncovered **5 critical bugs**, **4 duplication clusters**, and several medium-priority issues. This PR fixes all critical bugs, consolidates duplicated logic, and adds comprehensive test coverage (334 tests).

---

### Bug 1: Week Key Format Mismatch — Weekly Limits Never Enforced (CRITICAL)

**File:** `utils/subscriptionValidation.ts`
**Severity:** Critical — subscription weekly call limits were completely unenforced

**What was happening:**
Three methods in `SubscriptionValidationService` generate week-based map keys to track calls per week. They were using **two different key formats**, so lookups between them always returned 0:

| Method | Key Format | Example |
|--------|-----------|---------|
| `groupSlotsByWeek()` (proposed calls) | `YYYY-MM-DD` via `.split("T")[0]` | `"2025-01-05"` |
| `groupAppointmentsByWeek()` (existing calls) | Full ISO via `.toISOString()` | `"2025-01-05T00:00:00.000Z"` |
| `generateWeeklyInfo()` (lookup) | Full ISO via `.toISOString()` | `"2025-01-05T00:00:00.000Z"` |

When `generateWeeklyInfo()` tried to look up proposed calls with `proposedCalls.get(weekKey)`, the full ISO key `"2025-01-05T00:00:00.000Z"` never matched the truncated key `"2025-01-05"` stored by `groupSlotsByWeek()`. Result: `proposedCallCount` was always `0`.

**Impact:**
A user with a plan allowing 2 calls/week could submit 5 calls in the same week and the system would accept all of them. The weekly limit validation existed in code but was effectively dead because it never had real proposed-call data to check against.

**Fix (`subscriptionValidation.ts:273`):**
Removed `.split("T")[0]` from `groupSlotsByWeek()`, making it return `weekStart.toISOString()` — the same full ISO format used by the other two methods. One-line change that restores the entire weekly limit enforcement pipeline.

---

### Bug 2: Appointment Over-Counting — Existing Calls Doubled (CRITICAL)

**File:** `utils/subscriptionValidation.ts`
**Severity:** Critical — falsely blocked valid bookings

**What was happening:**
`groupAppointmentsByWeek()` iterated over each **slot** in an appointment and incremented the weekly call counter by 1 for each slot:

```
Original code (pseudocode):
for each appointment:
  for each slot in appointment:
    weeklyCount[slot.week] += 1     // ← BUG: counts slots, not appointments
```

All sessions in this system are composed of 30-minute slots. A standard 1-hour consultation = 2 slots. This code counted it as **2 calls** instead of 1.

**Impact:**
A subscription plan with 2 calls/week would appear "full" after scheduling just 1 call (because the single 1-hour appointment consumed both slots). Users were blocked from scheduling their legitimate second call of the week. For 1.5-hour sessions (3 slots), one appointment consumed all 3 weekly call slots on a plan allowing 3 calls/week.

**Fix (`subscriptionValidation.ts:234-243`):**
Rewrote the method to count **one call per appointment** instead of one call per slot. For each appointment, it finds the earliest slot's start time to determine which week the call belongs to, then increments that week's counter by exactly 1.

---

### Bug 3: Weekly Limit Validation — Mathematically Impossible Check (CRITICAL)

**File:** `utils/subscriptionValidation.ts`
**Severity:** Critical — validation code was dead code

**What was happening:**
`validateWeeklyLimits()` computed the total calls per week using this formula:

```
Original code:
totalCallsForWeek = week.maxCalls - week.availableSlots
if (totalCallsForWeek > week.maxCalls) → ERROR
```

But `availableSlots` was computed in `generateWeeklyInfo()` as:

```
availableSlots = Math.max(0, callsPerWeek - totalCalls)
```

Substituting: `totalCallsForWeek = maxCalls - Math.max(0, maxCalls - totalCalls)`. This can never exceed `maxCalls` because `Math.max(0, ...)` is always ≥ 0, so `maxCalls - (≥0)` is always ≤ `maxCalls`. The `if (totalCallsForWeek > week.maxCalls)` condition was **mathematically impossible to trigger**.

**Impact:**
Even if Bug 1 were fixed and proposed calls were correctly counted, this validation would still never reject an over-limit submission. The error messages it was supposed to produce (e.g., "Week of Jan 6 exceeds call limit") were unreachable code.

**Fix (`subscriptionValidation.ts:402-412`):**
Added `proposedCalls` as a tracked field in `WeeklyCallInfo`. Changed the check to `totalCallsForWeek = week.existingCalls + week.proposedCalls`, then compared that directly against `week.maxCalls`. This is a straightforward sum that correctly catches violations.

---

### Bug 4: Warnings Silently Dropped (MEDIUM)

**File:** `utils/subscriptionValidation.ts`
**Severity:** Medium — UI information loss

**What was happening:**
In `validateSubscriptionSlots()`, warnings from `validateWeeklyLimits()` (e.g., "Week of Jan 6 is fully booked. 2/2 calls scheduled.") were only pushed to the result inside the `if (!weeklyValidation.isValid)` block:

```
Original code:
if (!weeklyValidation.isValid) {
  result.isValid = false;
  result.errors.push(...weeklyValidation.errors);
  result.warnings.push(...weeklyValidation.warnings);  // ← only reached on error
}
```

Warnings are produced for **fully booked weeks** — weeks where `existingCalls >= maxCalls` and `proposedCalls === 0`. This is an informational state, not an error state. Since there are no errors in this case, `weeklyValidation.isValid` is `true`, and the warnings are never added to the result.

**Impact:**
The `SubscriptionValidationDisplay` component could never show "fully booked" indicators for weeks that were at capacity. Users had no visibility into which weeks were full vs. which still had capacity.

**Fix (`subscriptionValidation.ts:149-156`):**
Moved `result.warnings.push(...)` **outside** the `if (!isValid)` block so warnings are always propagated regardless of whether errors exist.

---

### Bug 5: Circular Slot Duration Calculation (MEDIUM)

**File:** `utils/slotAllocation/SlotValidationService.ts`
**Severity:** Medium — produced wrong values for edge cases

**What was happening:**
The conflict check in `validate()` calculated slot duration as:

```
Original code:
slotDurationMinutes = Math.round((sessionDurationInHours * 60) / slots.length)
```

This divides the total session duration by the number of slots provided. For correct inputs (e.g., 1-hour session with 2 slots → 30 minutes), it produces the right answer. But the calculation is **circular** — it assumes the input is correct to produce the duration, which is then used to validate the input.

The real problem: this conflict check runs **before** event-specific validators that catch slot count mismatches. So if a user submits 3 slots for a 1-hour session (should be 2), the duration computes to 20 minutes, and the conflict check uses 20-minute windows — potentially missing real conflicts with adjacent bookings.

**Impact:**
Incorrect slot counts could pass the conflict check with wrong duration windows, only to be caught later by event-specific validators. In the worst case, a malformed request could pass conflict checking entirely (no overlaps found with 20-minute windows) and then fail later, leaving inconsistent state.

**Fix (`SlotValidationService.ts:73`):**
Replaced the calculation with `const slotDurationMinutes = 30` — all slots in this system are uniformly 30 minutes. This is a known architectural invariant. The conflict check now always uses the correct 30-minute window regardless of input.

---

### Bug 6: getSlotStatus Booking Detection (MEDIUM)

**File:** `calendarUtils.ts`
**Severity:** Medium — incorrect calendar UI state

**What was happening:**
`getSlotStatus()` determined whether a time slot was "booked" or "partially booked" based on an **exact time match**:

```
Original code:
if (slotStart === apptStart && slotEnd === apptEnd) {
  isBooked = true;
} else {
  isPartiallyBooked = true;
}
```

A 30-minute calendar interval (e.g., 10:00–10:30) that falls **within** a 60-minute appointment (10:00–11:00) would fail the exact match and be marked as "partially booked" instead of "booked".

**Impact:**
Calendar slots that were fully occupied showed ambiguous "partially booked" styling instead of clear "booked" styling. Users could see a slot appear partially available when it was actually completely taken.

**Fix (`calendarUtils.ts:308-311`):**
Changed the check to containment: `slotStart >= apptStart && slotEnd <= apptEnd`. If the calendar interval is fully contained within an appointment's time range, it's "booked". Only true partial overlaps (slot extends beyond the appointment boundary) are "partially booked".

---

### Bug 7: Timezone Mismatch in Calendar Display (MEDIUM)

**File:** `calendarUtils.ts`
**Severity:** Medium — timezone-dependent display errors

**What was happening:**
`mapWeeklySlots()` projected weekly availability onto calendar dates using **local time** methods:

```
Original code:
slotStartTime.setHours(startTime.getHours(), startTime.getMinutes(), 0, 0);
```

But the server-side `SlotValidationService.validateMatchesSchedule()` uses **UTC methods** (`getUTCHours()`) to check if submitted slots fall within consultant availability. A consultant's 9 AM UTC availability would display at different times depending on the user's browser timezone, and when the user selected the displayed slot, the server might reject it because the UTC hours didn't match.

**Impact:**
In non-UTC timezones, slots could appear available in the calendar but fail server validation on submission, or vice versa. For example, in UTC+5:30 (IST), a 9 AM UTC slot displays at 2:30 PM local time. The user selects 2:30 PM, but if `setHours()` stored it as 2:30 PM UTC instead of 9:00 AM UTC, the server rejects it as "outside availability hours".

**Fix (`calendarUtils.ts:143-151`):**
Changed `.setHours()` to `.setUTCHours()` and `.getHours()` to `.getUTCHours()`, matching the server-side UTC convention. Calendar display and server validation now use the same time reference.

---

### Week Counting: 4 implementations → 1

Four files had independent implementations of "count Sunday-to-Saturday weeks" and "get Sunday start of week":

| File | Before | After |
|------|--------|-------|
| `SlotCalculationService.ts` | `countWeeks()`, `startOfWeekSunday()` | **Canonical source** (unchanged) |
| `calendarUtils.ts` | `countSundayWeeksInclusive()`, `startOfWeekSunday()` | Re-exports from SlotCalculationService |
| `allocationAlgorithms.ts` | Private `getWeekStart()` method | Calls `SlotCalculationService.startOfWeekSunday()` |
| `subscriptionValidation.ts` | `date-fns startOfWeek()` with options | Uses `SlotCalculationService.startOfWeekSunday()` |

### calculateRequiredSlots: 2 implementations → 1

`calendarUtils.ts` had a standalone `calculateRequiredSlots()` with different parameter signatures than `SlotCalculationService.calculateRequiredSlots()`. Replaced with a thin wrapper that adapts the legacy parameter signature and delegates to the canonical implementation.

---

## Test Coverage — 334 tests across 7 suites

| Suite | Tests | Key Coverage |
|-------|-------|-------------|
| `slotCalculationService` | 58 | Week counting, slot math, Sunday boundaries, edge cases |
| `calendarUtils` | 60 | Slot mapping, status detection, timezone handling, validation |
| `allocationAlgorithms` | 32 | Pre-allocation, slot distribution, session duration handling |
| `subscriptionValidation` | 32 | Weekly limits, call counting, period validation, warnings |
| `slotValidationService` | 32 | Schedule matching, conflict detection, slot consecutiveness |
| `slotAllocationService` | 59 | Mode routing, manual/auto/requested allocation, CRUD ops |
| `rescheduleCancel` | 60 | Reschedule API (24h policy, partial/full), Cancel API (notifications, waitlist), cleanup |

---

## Test plan

- [x] All 334 tests pass (`npx jest __tests__/booking-algorithm/`)
- [x] TypeScript compilation clean (`npx tsc --noEmit`)
- [ ] Manual: test subscription allocation API with slots exceeding weekly limits — should now be rejected
- [ ] Manual: test auto-allocation for 1-hour and 1.5-hour sessions — verify correct slot counts
- [ ] Manual: test calendar display in non-UTC timezone — verify slots appear at correct times
- [ ] Manual: end-to-end booking flow for consultation, subscription, webinar, and class

---

### Bug 8: Consecutive Slots Not Merged Before Duration Breakdown (MEDIUM)

**File:** `ConsultationPricingToggle.tsx`
**Severity:** Medium — bookable time slots missing for consultees

**What was happening:**
`breakDownSlotsByDuration()` received raw 30-minute availability slots without merging consecutive ones first. A consultant with two adjacent 30-minute slots (e.g., 10:00–10:30 and 10:30–11:00) should offer a 1-hour session, but the breakdown function saw them as isolated 30-minute slots and couldn't form the longer session.

**Impact:**
Consultees saw fewer bookable options than should exist. A consultant with 2 hours of contiguous availability (four 30-min slots) only showed 30-minute booking options instead of also offering 1-hour and 1.5-hour sessions.

**Fix:**
Added `mergeConsecutiveSlots()` call before `breakDownSlotsByDuration()`. Consecutive 30-minute slots are now merged into their full contiguous ranges before being broken down by the selected session duration.

---

### Bug 9: Cancelled/Rejected/Expired Appointments Show as "Booked" in Calendar (HIGH)

**File:** `useCalendarData.ts`
**Severity:** High — ghost bookings blocking real availability

**What was happening:**
`fetchExistingAppointments()` calls `/api/slots/appointments` without any status filter. The API returns **all** appointments regardless of status (CANCELLED, REJECTED, EXPIRED, etc.). The calendar's fallback logic in `getSlotStatusForInterval()` then marks any time interval overlapping with an appointment as "Booked" (gray), even if that appointment was cancelled or rejected.

**Root cause chain:**
1. `AllocationService.fetchAppointments()` passes no status params
2. The API route at `/api/slots/appointments` accepts status params (`consultationStatus`, `subscriptionStatus`, etc.) but **never applies them** in the Prisma WHERE clause — they are validated then silently ignored
3. The calendar fallback at `useCalendarData.ts:577` marks any overlapping appointment as booked: `if (overlappingAppointments.length > 0) { isBookedForDisplay = true; }`

**Impact:**
After a consultation is cancelled or a request is rejected, the freed time slot continues to appear as "Booked" in the Manage Timings calendar. Consultants cannot allocate new events to those slots without manually verifying they're actually free. This compounds over time as more cancellations accumulate.

**Fix:**
Added client-side filtering in `fetchExistingAppointments` after the existing validation step. Appointments are excluded if:
- Consultation/Subscription: `requestStatus` ∈ {`REJECTED`, `CANCELLED`, `EXPIRED`}
- Webinar/Class: `status` = `CANCELLED`

All other statuses (PENDING, APPROVED, APPROVED_PENDING_PAYMENT, SCHEDULED, COMPLETED) remain visible since they represent active or historical commitments.

---

### Bug 10: Multi-Session Events Only Show First Session as "This Event" (MEDIUM)

**File:** `useCalendarData.ts`
**Severity:** Medium — misleading calendar display for subscriptions/classes

**What was happening:**
`fetchEventSlots()` only processed `data[0].slotsOfAppointment`, discarding slots from all other appointments:

```
Original code:
if (data && data.length > 0 && data[0].slotsOfAppointment?.length > 0) {
  const slots = data[0].slotsOfAppointment.flatMap(...)
```

For a subscription with 4 weekly appointments or a class with multiple sessions, the API returns an array of appointments. Only the first appointment's slots were marked as "This Event" (black); the remaining sessions appeared as generic "Booked" (gray), indistinguishable from other consultants' bookings.

**Impact:**
When a consultant opens the Manage Timings modal for a multi-session subscription or class, they only see 1 session highlighted as belonging to that event. The other sessions blend in with unrelated bookings, making it impossible to visually verify the full allocation.

**Fix:**
Changed to process all appointments:
```
if (data && Array.isArray(data) && data.length > 0) {
  const slots = activeData.flatMap(appointment =>
    (appointment.slotsOfAppointment || []).flatMap(...)
  )
```

---

### Bug 11: Cancelled Event Appointments Shown as "This Event" (MEDIUM)

**File:** `useCalendarData.ts`
**Severity:** Medium — stale event markers after cancellation

**What was happening:**
`fetchEventSlots()` fetched all appointments for a specific event (by subscriptionId, classId, etc.) without filtering by status. If an event had cancelled or rejected appointments from a previous scheduling attempt, those old slots still displayed as "This Event" (black).

**Impact:**
After rescheduling an event (cancel old slots + allocate new ones), the calendar showed both the old cancelled slots and the new active slots in black "This Event" styling, doubling the apparent allocation.

**Fix:**
Added the same inactive-status filtering used in Bug 9 to the `fetchEventSlots` pipeline. Cancelled/rejected/expired appointments are excluded before processing slots into "This Event" markers.

---

### Bug 12: Stale Closures in Calendar Data Fetch Callbacks (LOW)

**File:** `useCalendarData.ts`
**Severity:** Low — affects mode/date switching edge cases

**What was happening:**
Both `fetchAvailabilitySlots` and `fetchExistingAppointments` are `useCallback` hooks that reference `mode`, `allowedStart`, and `allowedEnd` inside their bodies, but these variables were missing from the dependency arrays:

```
// fetchAvailabilitySlots:
const startDate = mode === "allocate" && allowedStart ? allowedStart : ...
}, [consultantId, toast, view, currentDate]);  // ← missing mode, allowedStart, allowedEnd

// fetchExistingAppointments:
const startDate = mode === "allocate" && allowedStart ? allowedStart : ...
}, [consultantId, toast, view, currentDate]);  // ← same issue
```

**Impact:**
When switching between "allocate" and "view" modes, or when the subscription scheduling period changes, the callbacks continued using stale closure values. The calendar would fetch data for the wrong date range until the next re-render triggered by an unrelated state change.

**Fix:**
Added `mode`, `allowedStart`, `allowedEnd` to both `useCallback` dependency arrays.

---

### Week Navigation for Custom Availability (Explore Page)

**Files:** `ConsultantAvailability.tsx`, `CustomAvailability.tsx`

Custom availability on the consultant explore page was locked to the current 7-day window. Consultees had no way to view availability beyond the current week.

**Changes:**
- Added `weekOffset` state and `handlePrevWeek`/`handleNextWeek` callbacks to `ConsultantAvailability`
- Updated the availability fetch effect to compute the 7-day window as `addDays(today, weekOffset * 7)`
- Added left/right chevron navigation arrows flanking the 7-column grid in `CustomAvailability`
- Left arrow is disabled at `weekOffset === 0` (can't navigate to past weeks)
- Each navigation triggers a fresh fetch for the new date range

---

## Updated Test Coverage — 538 tests across 14 suites

| Suite | Tests | Key Coverage |
|-------|-------|-------------|
| `slotCalculationService` | 58 | Week counting, slot math, Sunday boundaries, edge cases |
| `calendarUtils` | 60 | Slot mapping, status detection, timezone handling, validation |
| `allocationAlgorithms` | 32 | Pre-allocation, slot distribution, session duration handling |
| `subscriptionValidation` | 32 | Weekly limits, call counting, period validation, warnings |
| `slotValidationService` | 32 | Schedule matching, conflict detection, slot consecutiveness |
| `slotAllocationService` | 59 | Mode routing, manual/auto/requested allocation, CRUD ops |
| `rescheduleCancel` | 60 | Reschedule API (24h policy, partial/full), Cancel API (notifications, waitlist), cleanup |
| `+ 7 additional suites` | 205 | Stream chat, scheduling helpers, booking mock data, integration |

- [x] All 538 tests pass (`npm run test`)


lib/prisma.ts
teetangh, line 1: Hey @shubham79a, please dont comment out this. Will be helpful in the long run to optimize slow queries later on

shubham79a: Yes, but this causes problem in debugging. Logs got filled with slow query message.

teetangh:
@shubham79a, thanks for the booking algorithm fixes. lets fix the build tests and keep the resculding refacor and fixes as part of a different PR as i think that would be beyond scope for this pr. Also, gemini review on both prs would be more focused and would help catch blindspots from claude

shubham79a (approved):

Issue #446, opened 2026-02-19
Architecture Review: Familiarise Web Platform


teetangh:
**Overall Verdict: 7.5/10 — Impressively ambitious, mostly well-executed, with a few nightmare-grade issues**

For a pre-launch platform built by a 2-3 person team, this is genuinely impressive in scope and thoughtfulness. But there are specific areas that will cause real pain at scale. I'll be direct about each.

---

### What I Like

- **Polymorphic Appointment model is clever.** The `Appointment` model with nullable FK columns (`consultationId`, `subscriptionId`, `webinarId`, `classId`) and an `AppointmentsType` enum is a pragmatic choice. It avoids the complexity of table-per-type inheritance while keeping queries simple. The `SlotOfAppointment` many-to-many with `User` is the right call for group events.

- **Cancellation tracking is thorough.** Both `Consultation` and `Subscription` have `cancellationReason`, `cancellationNotes`, `cancelledAt`, `cancelledBy`. This is production-grade audit trail thinking.

- **Index strategy is generally solid.** Composite indexes like `@@index([requestedById, requestStatus])`, `@@index([status, holdUntil])`, and `@@index([isTentative, startsAt, endsAt])` show someone who understands query patterns.

- **The payout system is surprisingly mature** — `ConsultantEarnings`, `Payout`, `PayoutAccount`, `Invoice` with proper status enums, idempotency keys, and batch processing. This is ahead of most startups at this stage.

### What I Don't Like

1. **The `User` model is a god object (84 lines, 30+ fields, 20+ relations).**

   ```
   User → consultantProfile, consulteeProfile, staffProfile, adminProfile,
          slotsOfAppointment, payments, feedbacks, supportTickets, accounts,
          sessions, members, referralCode, referral, referralCredits,
          reportsSubmitted, reportsReceived, moderationActions,
          workExperiences, certifications, education, cookiePreferences,
          notificationPreferences, waitlist
   ```

   Every query that touches `User` risks pulling in this entire relation graph. At 10K+ users with eager loading mistakes, this becomes a performance cliff. The `consultantProfileId` and `consulteeProfileId` stored directly on `User` (with `@unique`) means a user can only ever be ONE consultant and ONE consultee. That's fine semantically but creates tight coupling.

2. **The `SlotOfAvailabilityWeekly` time representation is a nightmare waiting to happen.**

   ```prisma
   model SlotOfAvailabilityWeekly {
[...]

PR #462, opened 2026-03-02
fix: weekly availability DateTime→Int + autoAllocate distributed lock

teetangh:
### The core problem: DateTime for time-of-day (weekly slots)

`SlotOfAvailabilityWeekly` stored a consultant's recurring weekly availability (e.g., "Mondays 9:00–11:00 AM UTC") using `DateTime @db.Timestamptz()` fields anchored to `1970-01-05`. The date portion was completely meaningless — only the time mattered. This created a cascade of problems:

- **Every consumer had to "know the trick"**: extract `getUTCHours()` and `getUTCMinutes()` from a Date object, ignoring the date. New code touching weekly slots inevitably introduced bugs by treating these as real dates.
- **170+ lines of workaround logic** in `SlotValidationService.validateMatchesSchedule()` existed solely to wrestle DateTime objects into time-of-day comparisons.
- **The 1970-01-05 convention was undocumented tribal knowledge.** Anyone querying the table directly (Prisma Studio, SQL, debugging) would see dates from 1970 and be confused.

### Why Int instead of `@db.Time`

The architecture review (#446, Issue #6) suggested `@db.Time` as the minimum fix. We went further with `Int` (minutes since midnight UTC, 0–1439) because **`@db.Time` doesn't actually solve the root cause in a Prisma + JavaScript stack**:

| | `DateTime @db.Timestamptz` (old) | `DateTime @db.Time` | `Int` (our fix) |
|---|---|---|---|
| PostgreSQL storage | Timestamp with TZ | TIME type | Integer |
| Prisma JS client returns | `Date` object | **`Date` object** | `number` |
| Date portion present? | Yes (1970-01-05) | **Yes (1970-01-01)** | No |
| Consumer must extract time? | Yes | **Yes** | No — it IS the time |
| `540` obviously means 9:00 AM? | N/A | N/A | Yes |
| DB-level range constraint? | No | Partial | `CHECK (0 ≤ x < 1440)` |
| Simple arithmetic comparison? | No | No | `startTimeUtc < endTimeUtc` |

**The critical insight**: Prisma maps PostgreSQL `TIME` to JavaScript `Date`. You'd still get `new Date("1970-01-01T09:00:00.000Z")` and still need `.getUTCHours()` extraction everywhere. The `@db.Time` approach moves the meaningless date from `1970-01-05` to `1970-01-01` — same problem, different epoch anchor.

With `Int`, the value `540` is self-documenting: 540 minutes ÷ 60 = 9 hours = 9:00 AM UTC. No Date objects, no extraction, no confusion.

### Why custom slots keep `DateTime @db.Timestamptz`

Custom slots (`SlotOfAvailabilityCustom`) represent **specific calendar dates** — "March 15, 2026, 2:00–4:00 PM UTC". Unlike weekly slots where the date was fake (1970-01-05 anchor), custom slots need real dates — the date portion IS the data. `@db.Timestamptz` is the semantically correct PostgreSQL type for "a specific moment in time." We only renamed the fields (`availabilityStartsAt` → `startsAt`, `availabilityEndsAt` → `endsAt`) for naming consistency, without changing the type.

### The secondary fix: autoAllocate() race condition

`autoAllocate()` discovers and assigns available slots inside a Prisma transaction, but Prisma uses `READ COMMITTED` isolation — two concurrent auto-allocations can both read the same slot as "available" and both write to it. The existing `lockSlotBooking()` was used in checkout and request-approval flows, but **not** in auto-allocation.

Added a consultant-level Redis distributed lock (`auto-allocate:{consultantProfileId}`) acquired before the transaction and released in a `finally` block. This serializes concurrent auto-allocations for the same consultant, preventing double-booking without affecting different consultants' allocations.

### Post-review follow-ups (added in subsequent commits)

**PR #462 feedback fixes:**
- Added `@db.SmallInt` to `startTimeUtc`/`endTimeUtc` (2 bytes vs 4; range 0–1439 fits). The `CHECK` constraint is deferred to migration time.
- Fixed overnight split when `endTime = "00:00"`: hardcoded `endTimeUtc = 1439` instead of the computed `endMinutes = 0`, which caused `startTimeUtc > endTimeUtc` rejection for UTC-timezone consultants.
- Bumped `lockAutoAllocate` TTL from 120 s to 130 s to cover Redlock 1% drift factor (~128.7 s effective, 8.7 s buffer over the 120 s transaction timeout).

**CUIDv2 + checkout error handling:**
- Updated `eventIdSchema` CUID regex from `/^c[a-z0-9]{24}$/i` (CUIDv1-only, 25 chars) to `/^[a-z][a-z0-9]{23,24}$/` to accept both CUIDv1 (25 chars) and CUIDv2 (24 chars). Prisma 7.3 generates CUIDv2 by default for `Subscription`, `Webinar`, and `Class` models — the old regex silently rejected every auto-generated ID.
- Added `ZodError` handling in the checkout API to return structured 400 responses (instead of 500) with `AVAILABILITY_ERROR` classification.
- Fixed double-toast issue in consultation and unified checkout by returning after `handleApiError` instead of re-throwing.
- Added `handleDecline` for consultation requests in `RequestSlotAllocationTab`; improved table layout with fixed column widths and stacked action buttons.

**Calendar performance (O(1) slot lookup):**
- Replaced per-render O(n) scans in `getSlotStatusForInterval` with a `useMemo`-computed `slotStatusMap` (keyed by slot start time). With up to 672 slots rendered per week view (48 × 14), this eliminates ~336× linear scans per render.
- Added `selectedSlotsSet` and `eventSlotsSet` (`Set<number>`) for O(1) selection and event-slot matching.
- Split `useEffect` in `useCalendarData` into date-independent (consultant profile + event slots) and date-dependent (availability + appointments) fetches — week navigation now fires 2 API calls instead of 4.
- Reduced CSS `transition-duration` from 150 ms to 75 ms for snappier slot clicks.

**Unscheduled event UI + checkout availability validation:**
- Added `UnscheduledEventCard` component: shows class/webinar events that exist in the DB but have no appointment slots yet, with a "Set Schedule" button that opens `EventTimingsCalendar`.
- Added `syntheticAppointments.ts`: builds minimal `TAppointment` shapes from `UnscheduledClass` / `UnscheduledWebinar` to satisfy the calendar's expected interface without a real appointment record.
- `AppointmentsTab.tsx`: renders unscheduled cards per type section; suppresses the "No X scheduled" empty state when unscheduled cards are present.
- `page.tsx`: fetches `/api/events/classes` and `/api/events/webinars` for the consultant; passes unscheduled items down to `AppointmentsTab`.
- Added availability-window validation in `checkout.ts` (step 0b): verifies the requested slot falls within a linked `SlotOfAvailabilityWeekly` or `SlotOfAvailabilityCustom` record before conflict checks run.
- Moved waitlist creation for full webinars outside the DB transaction — it was being rolled back on `throw`; handles both inner-catch and outer-catch paths.
- Allocate routes (class, consultation, webinar): return HTTP 400 (not 500) for validation errors prefixed with `"Validation failed:"` or `"Invalid slot count:"`.
- Extended `requiredSlots` memo in `useSlotAllocation.ts` to use `maxTotalCalls` for CLASS (was subscription-only) — fixes slot count calculation for class events.
- Fixed seed file `6a-create-appointments.ts`: now builds a `consultantUserMap` and connects the consultant user to every `SlotOfAppointment` via the `_SlotOfAppointmentToUser` M2M join table (was only connecting the consultee).

## Commits

| # | Hash | Summary |
|---|------|---------|
| 1 | `8d1897ee` | fix: replace DateTime with Int for weekly availability + add autoAllocate lock |
| 2 | `1643fb28` | refactor: rename SlotOfAvailabilityCustom fields for consistency |
| 3 | `4322d36e` | fix: PR #462 feedback — SmallInt schema, UTC midnight edge case, lock TTL drift |
| 4 | `6a6320f4` | docs: update E2E test prompts for schema rename + add new test scenarios |
| 5 | `3affca10` | fix: accept CUIDv2 in eventIdSchema + checkout error handling + request decline |
| 6 | `2222df1c` | perf: calendar slot lookup O(1) + split date-independent fetches |
| 7 | `f0536a68` | fix: unscheduled class/webinar UI + checkout availability validation + seed consultant linking |

## Test plan

- [x] `npx prisma generate` — schema compiles
- [x] `npx tsc --noEmit` — zero TypeScript errors across 62 file changes
- [x] `npm test` — 623/623 tests pass, 17/17 suites
- [x] E2E test agent-001 — Consultation + Subscription (manual + auto-allocate, checkout, reschedule, cancel, race condition)
- [x] E2E test agent-002 — Webinar + Class (manual allocate, checkout, weekly limit UI, reschedule, cancellation, availability CRUD)
- [x] E2E test agent-003 — Webinar + Class **auto-allocate** (`isAuto: true`): consecutive block search, weekly distribution, idempotency, impossible-block error, tight scheduling period error, `isAuto` vs `slots` precedence
- [ ] Clean DB + reseed with updated seed files
- [ ] Manual: create consultant with weekly availability via onboarding flow
- [ ] Manual: verify calendar renders correctly on consultant dashboard
- [ ] Manual: trigger auto-allocation, confirm lock acquisition in Redis logs
- [ ] Manual: verify explore page shows correct availability times


teetangh:
## Re: Comment #9 — `@db.SmallInt` on weekly time fields (commit 4322d36e)

Added `@db.SmallInt` to `startTimeUtc` and `endTimeUtc` on `SlotOfAvailabilityWeekly`.

**Why `Int` (minutes-since-midnight) instead of `timestamptz`:**

Weekly availability is a **recurring pattern**, not a specific point in time. A slot like "every Monday 22:00–23:59 UTC" has no date — it repeats forever. `timestamptz` requires a concrete date, so you'd need to anchor to a fake reference date (e.g. Jan 1 1900), then strip the date on every read and reconstruct on every comparison. That's misleading and wasteful.

Minutes-since-midnight (0–1439) is the natural representation for dateless time-of-day:

- **No fake dates** — `1320` means "22:00 UTC", period
- **Trivial arithmetic** — overlap checks are integer range comparisons (`startA < endB && startB < endA`)
- **Day-of-week is a separate enum** — `timestamptz` would also need to encode this via an arbitrary date
- **SmallInt (2 bytes)** vs timestamptz (8 bytes) — 4× smaller, no timezone/DST ambiguity since values are already UTC

The `CHECK (value BETWEEN 0 AND 1439)` constraint is deferred to the actual migration run. All write paths already enforce the range via Zod (`z.number().int().min(0).max(1439)`).

teetangh:
## Calendar Performance Optimization

### Changes (commit `2222df1c`)

**1. Precomputed `slotStatusMap`** (`useCalendarData.ts`)
- Added `visibleDates`, `parsedRawSlots`, `parsedAppointmentSlots` useMemos to pre-parse data once
- Added `slotStatusMap` useMemo that precomputes all visible cell statuses into a `Map<string, SlotStatusResult>`
- `getSlotStatusForInterval` is now an O(1) Map lookup (was 336× O(n) linear scans per render)

**2. `selectedSlotsSet` + `eventSlotsSet`** (`useSlotAllocation.ts` + `UnifiedCalendar.tsx`)
- `selectedSlotsSet`: `Set<number>` replaces `selectedSlots.some()` O(n) with `Set.has()` O(1)
- `eventSlotsSet`: `Set<number>` replaces `eventSlots.some()` O(n) with `Set.has()` O(1) in both `renderTimeCell` and `handleSlotClick`

**3. Split `useEffect`** (`useCalendarData.ts`)
- Date-independent effect (consultant details + event slots) — fires on dialog open only
- Date-dependent effect (availability + appointments) — fires on dialog open + week navigation
- Week navigation now triggers 2 API calls instead of 4

**4. CSS transition** (`UnifiedCalendar.tsx`)
- `duration-150` → `duration-75` for snappier perceived click response

### Verification
- `npx tsc --noEmit` passes with zero errors

teetangh:
## Slot building DRY refactor + overnight format alignment (`a35fb96a`)

### What changed

**Problem:** The onboarding form (`ConsultantPreferredScheduleForm.tsx`) had ~85 lines of inline slot-building logic duplicated from `utils/schedule/formatting.ts`, and the two flows produced *different* DB representations for overnight slots:

| Flow | Overnight storage (e.g. Mon 10 PM → Tue 2 AM) |
|------|-----------------------------------------------|
| Dashboard (`formatSlotsForApi`) | **1 record** — `MONDAY startTimeUtc=1320 endTimeUtc=120` (`end < start` signals overnight) |
| Onboarding (inline `useEffect`) | **2 records** — `MONDAY 1320–1109` + `TUESDAY 1110–120` |

**Fix — two parts:**

**1. Extracted shared utilities** into `utils/schedule/formatting.ts`:
- `buildWeeklySlotsForSave(slots, timezone): WeeklySlot[]` — used by the onboarding form (replaces the inline `useEffect`)
- `buildCustomSlotsForSave(slots, timezone): CustomSlot[]` — same
- `WeeklySlot` / `CustomSlot` interfaces moved to `utils/schedule/types.ts`

**2. Aligned overnight format** — `formatWeeklySlot` (dashboard path) now also emits two monotonic records for overnight slots. The midnight split formula `(midnightMinutes + 1439) % 1440` is the single source of truth for both flows.

### Why two records instead of one

Every record now satisfies `startTimeUtc ≤ endTimeUtc`. `SlotValidationService` already handles both formats (it has explicit `isOvernightAvail = (endMinutes <= startMinutes)` detection), but the two-record split avoids that special-case path entirely — simpler queries, no per-record cross-midnight detection needed.

### What is unchanged

- `utils/onboarding-server.ts` — same payload types, no changes
- `SettingsTab.tsx` — already consumed `formatSlotsForApi`; benefits from the overnight alignment automatically with no code changes
- `PUT /api/user/consultants/[id]/route.ts` — unchanged; `dateToMinuteUtc` conversions work correctly with the new two-record ISO strings
- All 87 booking algorithm tests pass

teetangh:
Chat GPT 5.3 Codex Senior review findings focused on booking/rescheduling correctness after the Booking Algorithm 7 refactor.

I reviewed the Prisma refactor (`SlotOfAvailabilityWeekly` DateTime→Int + renamed custom fields), booking APIs, allocation services, checkout path, dashboard/calendar integration, stream changes, and related tests.

Targeted suites I ran locally all pass:
- `__tests__/booking-algorithm/slotAllocationService.test.ts`
- `__tests__/booking-algorithm/unallocatedSlotExclusion.test.ts`
- `__tests__/stream/channel-actions.test.ts`
- `__tests__/stream/event-channel-actions.test.ts`
- `npx tsc --noEmit`

Despite that, I found production-risk issues:

## 1) [P1] Weekly day-of-week is persisted from local day, not UTC day
**Where**
- `utils/schedule/formatting.ts:157-215`
- `utils/schedule/formatting.ts:278-334`
- `app/api/user/consultants/[id]/route.ts:345-350`

**Problem**
`startDay/endDay` are currently derived from local day keys (`day.toUpperCase()`), while `startTimeUtc/endTimeUtc` are UTC minutes. For slots near local midnight, UTC day can differ from local day. This stores inconsistent `(day, minute)` pairs.

**Impact**
- Weekly availability can shift by one day for some timezones.
- Auto-allocation (`SlotAllocationService`) and availability display can diverge.
- Intermittent “slot outside availability” behavior depending on timezone/time-of-day.

**Proposed fix**
- Derive `startDay/endDay` from converted UTC datetimes (`new Date(startUTC).getUTCDay()` mapped to `DayOfWeek`), not from local key.
- Do the same in both `formatWeeklySlot()` and `buildWeeklySlotsForSave()`.
- Keep all weekly records internally consistent as UTC day + UTC minute.

---

## 2) [P1] Checkout availability-window validation does not verify slot ownership (consultant scope)
**Where**
- `lib/payments/operations/checkout.ts:452-470`

**Problem**
The new check validates against the provided availability row id, but does not verify that row belongs to the consultant for the selected plan/event.

**Impact**
A crafted request can pass a different consultant’s `slotOfAvailabilityWeeklyId`/`slotOfAvailabilityCustomId` and potentially bypass intended schedule constraints.

**Proposed fix**
- Resolve expected consultantProfileId from the selected plan/event.
- Query availability row with both `id` and `consultantProfileId` (or compare fetched row’s consultant id explicitly).
- Reject if mismatch before time-window checks.

---

## 3) [P1] `/api/slots/availability/[consultantId]` removes booked slots globally, not consultant-specific
**Where**
- `app/api/slots/availability/[consultantId]/route.ts:159-177`

**Problem**
`removeBookedSlots()` queries appointments by start times only. There is no consultant filter.

**Impact**
If Consultant A has a booked slot at time `T`, Consultant B can lose `T` from availability response too. This creates false “booked” states across consultants.

**Proposed fix**
- Include consultant scoping in booking query (e.g., `slotsOfAppointment.some.user.some.id = consultantUserId`) and occupied-status filtering.
- Pass consultant user id into `removeBookedSlots()`.

---

## 4) [P2] Weekly slot APIs currently accept inconsistent cross-day payloads, but overlap/consumers assume same-day model
**Where**
- `app/api/slots/availability/weekly/route.ts:125-151`
- `app/api/slots/availability/weekly/[id]/route.ts:85-113`
- `app/api/slots/availability/weekly/[id]/route.ts:205-238`

**Problem**
API accepts `startDay !== endDay` as long as `startTimeUtc < endTimeUtc`, but overlap checks only scope by `startDay`. Other consumers assume split same-day records.

**Impact**
Malformed weekly records can be created and interpreted inconsistently (display/allocation overlap drift).

**Proposed fix**
Pick one model and enforce it:
- If split-record model: enforce `startDay === endDay` in POST/PUT/PATCH.
- If true cross-day model: implement cross-day overlap + validation across both days and update all consumers accordingly.

---

## 5) [P2] PUT overlap checks rely on optional `body.consultantProfileId`
**Where**
- `app/api/slots/availability/weekly/[id]/route.ts:96`
- `app/api/slots/availability/custom/[id]/route.ts:64`

**Problem**
When `consultantProfileId` is omitted, overlap query scope becomes ambiguous (and can behave incorrectly depending on ORM undefined handling/caller behavior).

**Impact**
False overlap conflicts or missed conflicts on update.

**Proposed fix**
- Fetch current slot first.
- Use `currentSlot.consultantProfileId` as authoritative fallback for overlap scope.
- Do not depend on client-provided consultant id for conflict domain.

---

## 6) [P1 outside booking core, but in this PR] Stream sync reconciliation can remove unrelated channel memberships
**Where**
- `actions/stream/chat/event-channel.action.ts:482-490`
- `actions/stream/chat/event-channel.action.ts:505-507`

**Problem**
Reconciliation queries *all* channels where user is a member, then removes membership for any channel id not in expected event+dm set.

**Impact**
User can be removed from non-event channels (collab/support/team/manual channels) during force sync.

**Proposed fix**
- Restrict cleanup to managed namespaces (`dm-*`, `webinar-*`, `class-*`) or managed metadata flag.
- Never reconcile unrelated channel types with this job.

---

If you want, I can send a follow-up comment with concrete patch snippets for each fix path.

teetangh:
Chat GPT 5.3 Codex additional findings from deeper pass (new issues beyond my previous review comment).

## 1) [P0] Migration is destructive and not safely deployable on non-empty prod data
**Where**
- `prisma/migrations/20260302081213_rename_availability_slots_and_add_payment_fields/migration.sql:33-46`
- migration warnings at `:4-16`

**Problem**
This migration drops old availability columns and adds new required columns without backfill. On existing data, this is either data-loss or migration failure.

**Impact**
- Potential complete loss of existing consultant availability.
- Production deploy can fail on non-empty tables due required-not-null additions.

**Proposed fix**
- Use phased migration:
  1. Add nullable new columns.
  2. Backfill from old columns (`day/start/end` and custom starts/ends).
  3. Validate row counts and data parity.
  4. Enforce `NOT NULL`, switch reads/writes, then drop legacy columns.
- Keep migration idempotent and include explicit SQL backfill.

---

## 2) [P1] Auto-allocation lock contention is misclassified as 500 instead of 409
**Where**
- `utils/appointmentlock.ts:728-730`
- `utils/slotAllocation/SlotAllocationService.ts:117-120`

**Problem**
Lock function throws message: `"Another auto-allocation is in progress..."`.
Error classifier only maps to 409 when message contains `"lock"`/`"Lock"`.

**Impact**
Concurrent auto-allocation returns 500 (unknown error) instead of 409 conflict, causing wrong client retry behavior and noisy error monitoring.

**Proposed fix**
- Prefer typed error classes/codes (`LOCK_CONTENTION`) from lock layer.
- Or minimally update classifier to detect the exact contention phrase (`"auto-allocation is in progress"`).

---

## 3) [P1] Weekly checkout availability-window validation checks start only, not full event duration
**Where**
- `lib/payments/operations/checkout.ts:464-470`

**Problem**
For weekly availability ID, validation ensures slot start minute is inside `[startTimeUtc, endTimeUtc)`, but does **not** ensure `slotEnd` is inside the same window.

**Impact**
A longer booking can start inside availability and extend beyond it (e.g., 60-min booking starting near window end), bypassing intended availability bounds.

**Proposed fix**
- Compute `slotEndMinutes` and enforce full-range inclusion for same-day windows.
- For cross-day weekly windows (if retained), apply normalized overnight-range checks consistently.

---

## 4) [P2] Custom-slot range filters use containment instead of overlap in multiple APIs
**Where**
- `app/api/slots/availability/custom/route.ts:31-36`
- `app/api/slots/unallocated/custom/route.ts:56-61`
- `app/api/slots/unallocated/[consultantId]/route.ts:69-73`

**Problem**
Date range filters require `startsAt >= start` **and** `endsAt <= end` (full containment). Slots that overlap range edges are excluded.

**Impact**
Missing availability/unallocated entries when slots cross date-window boundaries; this causes inconsistent calendar behavior near boundaries.

**Proposed fix**
- Use canonical overlap predicate everywhere:
  - `startsAt < endRange && endsAt > startRange`

---

## 5) [P2] Stale Stream cleanup is capped at 100 channels (no pagination)
**Where**
- `actions/stream/chat/event-channel.action.ts:482-486`

**Problem**
Reconciliation fetches only `limit: 100` channels and cleans stale memberships from that subset.

**Impact**
Users with >100 memberships get partial cleanup; stale memberships remain silently.

**Proposed fix**
- Iterate with pagination/cursor until exhaustion, or use `offset` loop with sane cap + logging if cap reached.

teetangh:
Chat GPT 5.3 Codex Third-pass findings (additional, non-duplicative).

## 1) [P1] Midnight/overnight split truncates availability by 1 minute and can drop the last 30-min slot
**Where**
- `prisma/schema.prisma:699-701` (weekly minute domain constrained to 0-1439)
- `utils/schedule/formatting.ts:185-196`
- `utils/schedule/formatting.ts:307-318`
- `utils/schedule/formatting.ts:381-388`

**Problem**
Overnight split uses `midnight - 1 minute` / `1439` sentinels (23:59) for first segment end. With half-open interval logic and 30-min slots, this can produce a 29-minute terminal interval and lose exact midnight-boundary coverage.

**Impact**
- Slot windows near midnight can be under-allocated / non-bookable unexpectedly.
- Calendar vs validation behavior drifts on edge intervals.

**Proposed fix**
- Model end-minute as exclusive 0..1440 (allow `1440`), or keep cross-day records instead of `-1 minute` truncation.
- Remove `23:59` sentinel strategy and standardize interval semantics globally (`[start, end)`).

---

## 2) [P1] Weekly availability checks are semantically inconsistent between allocator and validator
**Where**
- Allocator check: `utils/slotAllocation/SlotAllocationService.ts:681-685`
- Validator check: `utils/slotAllocation/SlotValidationService.ts:355-358`

**Problem**
Allocator allows a candidate if `candidateMinute < endTimeUtc` (start-only check).
Validator requires full 30-min coverage `slotEndTimeMinutes <= availEndMinutes`.

**Impact**
Edge slots can be selected by auto-allocation but rejected by validation, causing avoidable allocation failures and brittle behavior.

**Proposed fix**
- Use one canonical weekly interval predicate in shared utility and call it from both allocation and validation.
- Ensure predicate is duration-aware (full 30-min block) and consistent with chosen interval model.

---

## 3) [P2] Unallocated weekly endpoint ignores `endDay` and has no overnight handling
**Where**
- `app/api/slots/unallocated/weekly/route.ts:85-117`
- `app/api/slots/unallocated/[consultantId]/route.ts:110-143`

**Problem**
Weekly instance generation derives start/end on the same date from minute fields, matching only `startDay`; `endDay` is ignored.

**Impact**
For any cross-day/legacy weekly rows, overlap checks and returned unallocated windows can be wrong.

**Proposed fix**
- Normalize all weekly rows to enforced same-day model (recommended), or
- Add explicit cross-day/endDay-aware instance generation and overlap checks.

---

## 4) [P2] Collaborator availability day filtering is server-local and custom-slot query misses overlapping ranges
**Where**
- `app/api/collaborators/[consultantProfileId]/availability/route.ts:42-46`
- `app/api/collaborators/[consultantProfileId]/availability/route.ts:63`

**Problem**
`dayStart/dayEnd` are built with `setHours()` on server local timezone, not explicit UTC/user timezone.
Custom slots query uses `startsAt` containment only (`gte/lte`) instead of overlap.

**Impact**
- Date-boundary drift depending on server TZ.
- Slots that overlap the day but start before dayStart are omitted.

**Proposed fix**
- Build day bounds in explicit timezone/UTC.
- Use overlap predicate for custom slots (`startsAt < dayEnd && endsAt > dayStart`).

teetangh:
Chat GPT 5.3 Codex Fourth-pass findings (security and tenant isolation).

## 1) [P0 Security] Availability write endpoints have no auth or ownership checks
**Where**
- `app/api/slots/availability/weekly/route.ts` (POST)
- `app/api/slots/availability/weekly/[id]/route.ts` (PUT/PATCH/DELETE)
- `app/api/slots/availability/custom/route.ts` (POST)
- `app/api/slots/availability/custom/[id]/route.ts` (PUT/PATCH/DELETE)

**Problem**
These routes accept caller-provided consultant IDs / slot IDs and perform writes without session or ownership enforcement.

**Impact**
Any unauthenticated or unauthorized caller can create/update/delete consultant availability if they know IDs.

**Proposed fix**
- Require authenticated session.
- Resolve consultantProfileId from session user server-side.
- Enforce ownership on slot mutation (`where: { id, consultantProfileId: sessionConsultantProfileId }`).
- Remove trust in client-supplied consultant profile ids for writes.

---

## 2) [P1 Privacy] Public availability GET payloads include consultant email
**Where**
- `app/api/slots/availability/weekly/route.ts:35-39`
- `app/api/slots/availability/custom/route.ts:50-53`

**Problem**
Responses include `consultantProfile.user.email` in standard availability reads.

**Impact**
PII exposure risk via public endpoint calls.

**Proposed fix**
- Remove email from these responses unless explicitly required for an authenticated admin-only path.
- Return only non-sensitive fields needed by UI.

---

## 3) [P1 Integrity] Update routes allow reassignment across consultants via body payload
**Where**
- Weekly: `app/api/slots/availability/weekly/[id]/route.ts:129-131` and `:258-260`
- Custom: `app/api/slots/availability/custom/[id]/route.ts:94-96` and `:189-191`

**Problem**
`consultantProfileId` from request body can reconnect slot ownership to a different consultant.

**Impact**
Cross-tenant data tampering: slot ownership can be moved if caller can hit endpoint.

**Proposed fix**
- Disallow consultant reassignment from these mutation routes.
- Keep consultant ownership immutable after creation, or enforce privileged admin-only flow.

teetangh:
Chat GPT 5.3 Codex Fifth-pass findings (availability API correctness).

## 1) [P1] `mapWeeklySlotToTiming` rewrites UTC day for near-midnight slots
**Where**
- `app/api/slots/availability/[consultantId]/route.ts:232-239`
- `app/api/slots/availability/[consultantId]/route.ts:257-260`

**Problem**
After timezone conversion, the code force-resets slot date to `userDate` via `setToUserDate()`. For UTC times that naturally land on the next/previous local day, this rewrites the timestamp to the wrong day.

**Repro (conceptual)**
- Weekly slot: `23:00 UTC`.
- User TZ: `Asia/Kolkata` (should be `04:30` next day).
- Current logic forces date back to requested day, yielding `04:30` on wrong day and shifted UTC output.

**Impact**
Displayed/returned slot times can shift by 24h around day boundaries, causing incorrect booking windows.

**Proposed fix**
- Remove date-forcing (`setToUserDate`) from weekly mapping.
- Build local occurrence directly from intended day + UTC minute with explicit timezone math, preserving real local date transitions.

---

## 2) [P2] Custom daily availability query misses slots that fully cover the day
**Where**
- `app/api/slots/availability/[consultantId]/route.ts:126-139`

**Problem**
Query includes:
- starts within day OR
- ends within day
but omits the enclosure case: `startsAt <= dayStart && endsAt >= dayEnd`.

**Impact**
Long custom slots spanning the full requested day are dropped from daily availability response.

**Proposed fix**
- Add third overlap clause for enclosure, or use canonical overlap predicate directly:
  - `startsAt < dayEnd && endsAt > dayStart`

teetangh:
## Booking Algorithm Review Checklist (Current State)

| Area | Item | Status | Evidence |
|---|---|---|---|
| Auth/Ownership | `custom/[id]` `PUT/PATCH/DELETE` enforce auth + owner checks | [x] Fixed | Endpoint now checks session and compares slot owner userId before mutation |
| Auth/Ownership | Block consultant reassignment on custom slot updates | [x] Fixed | `consultantProfileId` change is rejected on update paths |
| Weekly Overlap | Same-day vs overnight overlap gap (3rd OR clause) | [x] Fixed | Overlap helper includes overnight-starting-same-day branch |
| Weekly Validation | Overnight cross-day rule enforces real cross-midnight (`startTimeUtc > endTimeUtc`) | [x] Fixed | Validation now rejects invalid >24h-style cross-day ranges |
| Allocation Correctness | Overnight candidate overflow near midnight (e.g. `23:50→00:20`) | [x] Fixed | Allocation now validates midnight overflow against next-day `endTimeUtc` |
| Availability API | `custom/[id]` PUT/PATCH invalid date payload handling | [x] Fixed | Invalid date strings now return 400 instead of surfacing as server error |
| Availability API | Consultant-scoped booked-slot filtering in `[consultantId]` availability GET | [x] Fixed | Booked slots are filtered with consultant user scoping |
| Lock Contention | Lock errors classified as HTTP 409 conflicts | [x] Fixed | Lock messages + classifier now map in-progress contention to `LOCK_CONTENTION` |
| Stream Channel Sync | Cleanup restricted to managed channel prefixes | [x] Fixed | Cleanup now filters to `consultation-`, `subscription-`, `webinar-`, `class-`, `trial-`, `dm-` |
| Stream Channel Sync | Stream membership pagination for >100 channels | [x] Fixed | Reconciliation pass now paginates `queryChannels` |

## Pending Items

No blocking or high-confidence pending bugs found in the reviewed booking/availability/lock/stream-sync scope of this PR.

## Worth Noting (Non-blocking Follow-up)

1. Add a targeted test for overnight edge-case fit at midnight boundary (`23:50→00:05` availability should reject a full 30-min slot starting `23:50`).
2. Add API tests for `PUT/PATCH /api/slots/availability/custom/[id]` invalid date payloads returning 400.
3. Add stream sync tests that explicitly cover paginated stale-channel cleanup with managed/unmanaged prefix separation.

## Validation Snapshot

- `tsc --noEmit`: pass
- Focused suites: booking allocation + stream channel sync tests: pass

teetangh:
Chat GPT 5.4 review pass focused on booking / rescheduling scope.

I found five issues in the current PR that should be addressed before merge:

1. `P1` `utils/slotAllocation/slotTimeUtils.ts`
Overnight weekly overlap detection still misses carry-over conflicts in the overnight branch of `buildWeeklyOverlapWhere()`.

Concrete failure case:
- existing slot: `MONDAY 22:00 -> TUESDAY 02:00`
- new slot: `TUESDAY 01:00 -> WEDNESDAY 03:00`

These overlap for one hour, but the current predicates do not match that shape, so both rows can be saved. Once that happens, downstream availability and allocation logic is working from contradictory source data.

Suggested fix:
- add an explicit carry-over overlap predicate for existing overnight slots whose `endDay === candidate.startDay`, or
- normalize both weekly slots into day-level segments before comparing.

2. `P1` `lib/payments/operations/checkout.ts`
The new checkout availability guard trusts `slotOfAvailabilityWeeklyId` / `slotOfAvailabilityCustomId` by primary key only. It never verifies that the slot belongs to the consultant attached to the plan being purchased.

That means a caller can use consultant A's checkout flow with consultant B's public availability slot ID, as long as the submitted timestamp fits B's slot window. Conflict detection is still scoped to consultant A, so this bypasses the intended protection instead of strengthening it.

Suggested fix:
- assert `avail.consultantProfileId === plan.consultantProfileId` before validating the time window, or
- thread the consultant profile id explicitly into the validation call.

3. `P1` `lib/payments/operations/checkout.ts`
The weekly availability guard is still same-day only. It requires:
- `slotDay === avail.startDay`
- `slotMinutes >= avail.startTimeUtc`
- `slotMinutes < avail.endTimeUtc`

That rejects legitimate overnight weekly bookings, including:
- late-night bookings in a slot where `endTimeUtc < startTimeUtc`
- post-midnight bookings for the `endDay` side of an overnight slot

The PR adds overnight support elsewhere, so checkout becomes the path where those bookings still fail.

Suggested fix:
- reuse the same overnight-aware matching logic already present in `SlotAllocationService.isWithinAvailability()` / `SlotValidationService.validateMatchesSchedule()`.

4. `P2` `utils/slotAllocation/SlotAllocationService.ts`
Weekly consultation / webinar auto-allocation is sorted by raw `startTimeUtc`, not by the actual next occurrence in the calendar.

Example:
- current time: Monday morning
- Monday 09:00 slot exists this week
- Tuesday 08:00 slot exists this week

Because the array is pre-sorted by clock time, Tuesday 08:00 is evaluated before Monday 09:00 and can win, even though Monday is earlier in real time. If the intended behavior is “first available consecutive block”, this is incorrect.

Suggested fix:
- compute candidate starts first and sort by `candidateStart.getTime()` before evaluating blocks.

5. `P2` `app/dashboard/consultant/[consultantId]/(features)/appointments/page.tsx`
The new appointments page filters webinars down to unscheduled items, but the class query passes all classes straight through as `unscheduledClasses`.

`AppointmentsTab` then renders `Set Schedule` cards for every class in that array, so already-scheduled classes appear twice:
- once as normal scheduled groups
- once again as unscheduled cards

Suggested fix:
- filter classes the same way webinars are filtered, only passing classes with no appointments.

I stayed within the PR scope and re-checked these against the changed code paths rather than restating the existing bot comments verbatim.

teetangh:
Chat Gpt 5.4 Follow-up review after the latest fixes and the new checklist.

I re-checked the booking/rescheduling paths on current head and found two more issues still inside this PR's scope.

1. **[P1] Overnight weekly slots are still misreported by the unallocated-slot APIs**  
   Files: `app/api/slots/unallocated/weekly/route.ts:95-123`, `app/api/slots/unallocated/[consultantId]/route.ts:121-149`

   Both routes still build `slotEnd` on the same UTC date as `slotStart` for weekly slots. That works for same-day availability, but for any overnight slot (`MONDAY 22:00 -> TUESDAY 02:00`) the computed `slotEnd` becomes earlier than `slotStart`. Once that happens, the overlap predicate `allocated.startsAt < slotEnd && allocated.endsAt > slotStart` stops matching real post-midnight bookings, so the route can report an occupied overnight weekly slot as unallocated.

   Concrete example:
   - weekly availability: `MONDAY 22:00 -> TUESDAY 02:00`
   - booked appointment: `TUESDAY 01:00 -> TUESDAY 01:30`
   - current generated range in these routes: `slotStart=Mon 22:00`, `slotEnd=Mon 02:00`
   - result: overlap check fails and the slot is surfaced as unallocated even though it is occupied.

   The fix should mirror the other overnight-aware paths: if `endDay !== startDay` or `endTimeUtc <= startTimeUtc`, roll `slotEnd` into the next UTC day before overlap checks and before serializing `slotEndTimeInUTC`.

2. **[P2] Scheduling-period validation still allows sessions to run past the end boundary**  
   Files: `utils/slotAllocation/SlotValidationService.ts:494-509`, `app/dashboard/consultant/[consultantId]/(features)/requests/components/RequestedSlotsDialog.tsx:97-105`

   `validateSchedulingPeriod()` only checks each slot's start timestamp against `[startDate, endDate]`. That is weaker than the rule stated in the error text: "All slots must be scheduled within this date range." For multi-slot sessions, the final 30-minute block can extend past `schedulingPeriodEndsAt` and still pass validation as long as its start is `<= endDate`.

   Concrete example:
   - scheduling period ends at `2026-06-30 10:00 UTC`
   - 1-hour session uses slots `09:30` and `10:00`
   - current validator accepts both starts
   - actual session runs until `10:30`, so part of the booking is outside the allowed period.

   The server validator should check full slot intervals, not only starts. For 30-minute slots that means validating `slot >= startDate` and `slot + 30m <= endDate` (or validating the entire allocation window once). The client-side mirror in `RequestedSlotsDialog` should follow the same rule so the UI and API stay aligned.

3. **Checklist extension suggestions**  
   `docs/booking/15-checklist.md` is useful, but I would extend it with these explicit guards because both are easy to regress:

   - In **Overnight / Cross-Midnight Handling** or **API Routes**: add a checkbox that both unallocated weekly routes roll `slotEnd` into the next UTC day for overnight slots.
   - In **Slot Allocation** or **Data Validation**: add a checkbox that scheduling-period validation checks the full session interval, not only slot start times.
   - In **Quick Smoke Test After Changes** add:
     - `Unallocated overnight: Mon 22:00->Tue 02:00 with a Tue 01:00 booking is excluded from both /api/slots/unallocated/weekly and /api/slots/unallocated/[consultantId]`
     - `Scheduling period boundary: a 1-hour session ending after schedulingPeriodEndsAt is rejected even if its last slot starts exactly at the end time`

Those are the only additional booking-path issues I could verify confidently on the current branch.

teetangh:
Chat Gpt 5.4  Further review on the latest head turned up three more booking-scope issues.

1. **[P1] `/api/user/consultants/[id]` still bypasses the ownership checks added to the dedicated availability CRUD routes**  
   File: `app/api/user/consultants/[id]/route.ts:217-377`

   The weekly/custom availability CRUD routes were hardened earlier in this PR, but the bulk consultant settings save route still only checks that *some* session exists. It never verifies `consultantProfile.userId === session.user.id` before overwriting the consultant profile and deleting/recreating availability rows. Any authenticated user who knows another consultant's profile ID can therefore replace that consultant's weekly/custom schedule through this route.

   This route is in the active booking path: `SettingsTab` still saves availability through `/api/user/consultants/${consultant.id}`. So the dedicated CRUD fixes are currently bypassable through the dashboard's own main save endpoint.

   Suggested fix: add the same ownership check here that the availability CRUD routes now use, and mirror it on `DELETE` as well.

2. **[P1] The same bulk save route bypasses the new server-side overlap / time-order protections**  
   File: `app/api/user/consultants/[id]/route.ts:335-377`

   After validating only the outer Zod shape, this route converts the submitted weekly/custom slots and writes them straight with `createMany()`. It does not call `validateWeeklySlotTimeOrder()`, `buildWeeklyOverlapWhere()`, or the custom overlap predicate that the dedicated availability endpoints now enforce. A direct request to this route can therefore persist overlapping weekly windows, overlapping custom windows, or otherwise invalid schedule combinations even though the individual availability APIs now reject them.

   In practice this re-opens the same source-of-truth corruption risk the PR is trying to close, just through a different endpoint.

   Suggested fix: either route dashboard saves through the hardened availability APIs or reuse the same server-side validators here before deleting/recreating rows.

3. **[P1] Overnight weekly slots are still blocked or truncated on the main save paths**  
   Files: `utils/timeSlotValidation.ts:40-75`, `utils/timeSlotValidation.ts:193-229`, `app/form/onboarding/components/ConsultantPreferredScheduleForm.tsx:23,195-199`, `app/dashboard/consultant/[consultantId]/(features)/settings/SettingsTab.tsx:582-616`, `utils/schedule/formatting.ts:160-203`, `utils/schedule/formatting.ts:290-318`

   There are two separate problems here:

   - The frontend validator still rejects genuine overnight slots (`22:00 -> 02:00`) with `"Overnight slots are not allowed"`. That means the backend's new overnight support is not reachable from onboarding/settings for the main UI flow.
   - Midnight-ending slots (`22:00 -> 00:00`) are allowed, but both save-path formatters normalize them to `23:59` / minute `1439`. That silently shrinks the slot by one minute and drops the final `23:30 -> 00:00` booking interval. More generally, the split-at-midnight path emits a `23:30 -> 23:59` fragment, which the calendar/validator logic will not treat as a full 30-minute bookable slot.

   So today the UI cannot create most overnight weekly availability at all, and for the one overnight shape it does allow, it stores a truncated window that loses the last bookable slot.

   Suggested fix: update the UI validator to allow overnight ranges in the same shapes the backend accepts, and stop encoding midnight boundaries as `23:59` / `1439`. The save-path split should preserve true half-hour slot boundaries across midnight rather than shaving a minute off the first segment.

4. **Checklist additions worth making**  
   `docs/booking/15-checklist.md` is strong, but I would add these because the current regressions sit outside the dedicated CRUD routes:

   - `Bulk consultant settings route (/api/user/consultants/[id]) enforces ownership and reuses the same availability validation as the CRUD routes`
   - `Main UI save paths (onboarding + settings) can create overnight weekly slots that backend booking logic accepts`
   - `Midnight boundary smoke test: 22:00->00:00 weekly availability still exposes the 23:30->00:00 slot`

Those are the additional issues I could verify confidently on the current branch.

teetangh:
Chat Gpt 5.4 Another pass on current head found two remaining availability-integrity gaps.

1. **[P1] Bulk consultant settings save still accepts overlapping custom availability slots**  
   File: `app/api/user/consultants/[id]/route.ts:419-445`

   The weekly branch in this route was hardened, but the custom branch still only checks `startsAt < endsAt` and then writes the whole submitted set with `createMany()`. There is no pairwise overlap check equivalent to the dedicated custom availability CRUD route. A direct request to `/api/user/consultants/[id]` can therefore still persist overlapping custom windows even though `/api/slots/availability/custom` now rejects them.

   That means the main dashboard save endpoint still has a second consistency hole: weekly overlaps are blocked, custom overlaps are not.

   Suggested fix: validate the submitted custom-slot set with the same canonical overlap predicate (`existing.startsAt < newEnd && existing.endsAt > newStart`) before deleting/recreating rows.

2. **[P1] Onboarding persistence is still behind the hardened availability validation rules**  
   Files: `app/api/form/onboarding/[id]/route.ts:27`, `utils/onboarding-server.ts:84-145`

   The onboarding PATCH route persists consultant availability through `processOnboardingData()` / `updateConsultantProfileAndRelations()`, but that code path still validates weekly slots only via `isValidTimeRange()` on `HH:MM` strings and validates custom slots only the same way. It never calls `validateWeeklySlotTimeOrder()`, never enforces `endDay` being the next day for overnight weekly slots, and never checks for pairwise overlaps in either weekly or custom submissions before `createMany()`.

   So even after the fixes in the dedicated availability APIs and in `/api/user/consultants/[id]`, onboarding can still write impossible weekly day combinations or overlapping weekly/custom availability if a caller bypasses the client-side form logic.

   Suggested fix: make onboarding reuse the same server-side validators as the dedicated availability routes / bulk settings route, rather than filtering only by `isValidTimeRange()`.

3. **Checklist additions**

   I would extend `docs/booking/15-checklist.md` with:

   - `Bulk consultant settings route rejects overlapping CUSTOM slot submissions, not only WEEKLY`
   - `Onboarding persistence path (/api/form/onboarding/[id] -> onboarding-server) enforces the same weekly/custom validation rules as the availability CRUD routes`

Those are the additional booking-path issues I could verify confidently on the current branch.

teetangh:
Next review pass found a remaining family of date-window / projection issues. These are related enough that I’m grouping them into one comment.

1. **[P1] Several custom-slot APIs still require full containment instead of using the canonical overlap predicate**  
   Files:
   - `app/api/slots/availability/[consultantId]/route.ts:123-139`
   - `app/api/slots/unallocated/custom/route.ts:56-68`
   - `app/api/slots/unallocated/[consultantId]/route.ts:68-73`
   - `app/api/collaborators/[consultantProfileId]/availability/route.ts:60-64`

   These routes still fetch custom availability with predicates like `startsAt >= rangeStart` and `endsAt <= rangeEnd` (or the two-branch start-inside/end-inside version), instead of the canonical overlap check `startsAt < rangeEnd && endsAt > rangeStart` that newer code already uses elsewhere.

   That means any custom slot that fully contains the requested day/range disappears from these APIs. Example:
   - custom slot: `2026-03-10T00:00Z -> 2026-03-12T00:00Z`
   - query range: `2026-03-11T00:00Z -> 2026-03-11T23:59:59Z`
   - expected: slot is returned because it overlaps the day
   - current result: slot is dropped because neither its start nor its end falls inside the query window.

   Impact:
   - public availability can show a consultant as unavailable on days that are actually covered by a long custom slot
   - unallocated custom APIs can undercount/report missing availability
   - collaborator overlay can hide valid co-host availability

   Suggested fix: make all custom-slot date-range queries use the same overlap predicate already used in `availability-with-allocation`.

2. **[P1] Collaborator availability overlay still misses overnight bookings that start the previous day**  
   File: `app/api/collaborators/[consultantProfileId]/availability/route.ts:72-95`

   `bookedSlots` are fetched with `startsAt >= dayStart && startsAt <= dayEnd`. That misses any booking whose occupied time overlaps the target day but whose `startsAt` was on the previous day.

   Concrete example:
   - booked slot: `2026-03-10 23:30 -> 2026-03-11 00:30`
   - collaborator overlay query for `2026-03-11`
   - expected: the co-host should appear booked for the first 30 minutes of the day
   - current result: the booking is omitted because it started on `2026-03-10`

   Suggested fix: use the same overlap window for booked slots (`startsAt < dayEnd && endsAt > dayStart`) rather than start-time containment.

3. **[P2] The simple public availability route still uses raw slot-row matching instead of occupancy-aware overlap filtering**  
   File: `app/api/slots/availability/[consultantId]/route.ts:159-192`

   `removeBookedSlots()` is still stale compared with the newer booking logic:
   - it queries appointments only by slot-row presence, with no parent-status filter, so any surviving slot rows from non-occupying states (`EXPIRED`, `REJECTED`, etc.) still hide availability
   - it removes slots only when an appointment `startsAt` exactly matches the returned slot's `slotStartTimeInUTC`

   That second point means partial occupancy inside a larger returned window is invisible. Example:
   - custom slot returned by this route: `09:00 -> 12:00`
   - booked appointment inside it: `10:00 -> 10:30`
   - current logic checks only whether some appointment starts exactly at `09:00`, so the slot is still returned as if fully open

   Even if this route is older than `availability-with-allocation`, it is still in the PR diff and still exposes booking-state logic that disagrees with the rest of the system.

   Suggested fix: either retire this route in favor of the allocation-aware one, or make it use the same occupancy policy + overlap semantics.

4. **Missing coverage / checklist additions**

   I would add these to `docs/booking/15-checklist.md` and to tests:

   - `Any custom-slot date-window query uses overlap semantics (startsAt < rangeEnd && endsAt > rangeStart), not full containment`
   - `Day-based booked-slot queries include overnight carry-in appointments from the previous day`
   - smoke test: long custom slot fully spanning the queried day appears in public availability, unallocated custom, and collaborator availability routes
   - smoke test: appointment starting previous day but ending after midnight blocks collaborator availability on the next day

I could not find focused tests covering these query-shape regressions on current head.

teetangh:
Next review pass found another cluster of booking-state issues that are worth fixing together.

## 1. P1: `cancel` still corrupts multi-appointment subscriptions/classes

Files:
- `app/api/appointments/[appointmentId]/cancel/route.ts` (`174-208`)
- `app/dashboard/consultee/[consulteeId]/(features)/appointments/Overview.tsx` (`236-239`, `406-409`)

The consultee overview still passes only `appointments?.[0]?.id` into the cancel action for subscriptions and classes, but the cancel route updates the parent `subscription` / `class` to `CANCELLED` and then deletes only that single `appointmentId`.

That leaves the rest of the appointment rows and slot rows behind under a cancelled parent event. In practice this creates contradictory state:
- the booking is shown as cancelled,
- some future sessions may still exist in the database,
- downstream queries that read raw appointment rows can keep surfacing stale sessions.

This is the same shape as the earlier partial-reschedule bug, but now on cancellation.

**Proposed fix**
- For `subscription` and `class`, cancel at the event scope, not the single appointment scope.
- Delete all appointments (or all future appointments, if that is the intended policy) for the parent event inside the same transaction.
- Add a regression test that cancels a 3-session subscription/class from the first appointment card and asserts that no sibling appointments remain.

## 2. P1: cancelling a webinar/class incorrectly notifies the waitlist that a spot opened

File:
- `app/api/appointments/[appointmentId]/cancel/route.ts` (`257-265`)

After cancelling a webinar or class, the route unconditionally calls `handleSlotOpening(... slotsAvailable: 1)`. But this route is cancelling the event itself for the organizer, not freeing one participant seat on a still-active event.

That means waitlisted users can receive a “spot available” email for an event that was just marked `CANCELLED` and had its appointment deleted.

**Proposed fix**
- Do not call `handleSlotOpening()` when the event itself is being cancelled.
- Reserve waitlist notifications for true capacity openings such as participant removal or capacity increase.
- Add a test that cancels a webinar/class and asserts no waitlist notification path is triggered.

## 3. P1: waitlisted/notified webinars and classes still do not appear on the consultee appointments dashboard

File:
- `app/api/dashboard/consultee/[consulteeId]/events/route.ts` (`113-121`, `169-178`)

The comments say webinars/classes are fetched when the user is registered “via waitlist or via appointment slots”, but the query only matches events where the user already has appointment slots.

The `include: { waitlist: { where: { userId } } }` block does not help, because the enclosing `where` clause already filtered out waitlist-only events.

So a user who is `WAITLISTED` or `NOTIFIED` for a webinar/class never sees that event in the dashboard, even though the UI already has logic to render `WAITLISTED` / `NOTIFIED` booking badges.

**Proposed fix**
- Change both webinar and class queries to `OR` between:
  - confirmed participation via appointment slots, and
  - active waitlist membership for this user.
- Keep the waitlist include scoped to the current user.
- Add a dashboard API test for a waitlist-only webinar/class response.

## 4. P1: approval-payment records are still orphaned from the appointment until webhook success, so the 48h expiry path cannot see them

Files:
- `lib/payments/operations/approval-payment.ts` (`90-148`, `230-289`)
- `app/api/cleanup/approval-payments/route.ts` (`116-150`)
- `scripts/appointments/cleanup-tentative-slots.ts` (`50-62`)

`createApprovalPaymentIntent()` creates a pending `Payment` row with no `appointmentId`, while `checkExistingPayment()` and the approval-expiration cleanup both look for payments by traversing `consultation.appointment.payment` / `subscription.appointments.payment`.

That means the pending approval-flow payment is invisible to the very logic that is supposed to deduplicate it and expire it after 48 hours.

Concrete failure mode:
- consultant approves a consultation/subscription,
- approval payment is created,
- payment remains pending and expires,
- `/api/cleanup/approval-payments` finds the `Payment`, but `payment.appointment?.consultation` / `subscription` is null, so no parent status is reverted.

The system then falls back to slower, inconsistent cleanup paths instead of the intended 48h approval-expiration path.

**Proposed fix**
- Link approval-flow payments to the existing tentative appointment immediately when the payment row is created.
- Or, if that is not possible, make dedupe/cleanup query by `consultationId` / `subscriptionId` directly instead of relying on `appointment.payment` before webhook success.
- Add an integration test for: approve request -> generate payment link -> expire payment -> parent status and tentative slots are cleaned by the intended approval-expiration job.

## 5. P1: a few remaining projection routes still use non-canonical date-window logic for custom slots / booked overlays

Files:
- `app/api/slots/availability/[consultantId]/route.ts` (`123-139`, `159-192`)
- `app/api/collaborators/[consultantProfileId]/availability/route.ts` (`60-75`)
- `app/api/slots/unallocated/custom/route.ts` (`56-68`)
- `app/api/slots/unallocated/[consultantId]/route.ts` (`68-73`)

There is still one family of routes that has not been brought onto the same overlap semantics as the newer allocation code:
- custom availability queries still use containment/start-or-end-inside filters instead of `startsAt < rangeEnd && endsAt > rangeStart`,
- collaborator booked-slot overlay still only matches `startsAt` inside the target day, so overnight carry-in bookings are missed,
- the simple public availability route still removes bookings by exact slot start matching rather than occupancy-aware overlap.

These are all variants of the same underlying problem: projection routes are still using row-shape assumptions instead of the canonical interval overlap policy.

**Proposed fix**
- Standardize all custom-slot date-window queries on the same overlap predicate.
- Standardize day-based booked-slot overlays on `startsAt < dayEnd && endsAt > dayStart`.
- Where possible, reuse the same occupancy-aware helper strategy the allocation path already uses.

## Checklist additions worth adding

- [ ] Cancelling a subscription/class removes all sibling appointments for that parent event, not only the clicked appointment row.
- [ ] Cancelling an entire webinar/class never triggers waitlist spot-open notifications.
- [ ] Consultee dashboard event queries include waitlist-only webinar/class memberships.
- [ ] Approval-flow pending payments are discoverable by both duplicate-payment checks and expiry cleanup before webhook success.
- [ ] All projection/date-window routes use canonical interval-overlap semantics, including overnight carry-in booked slots.

teetangh:
I re-verified the latest changes and the next review pass still found a waitlist / participant-state cluster worth fixing together.

## 1. P1: webinar waitlist availability still counts slot rows, not enrolled users

File:
- `lib/waitlist/slot-handler.ts` (`334-365`)

`checkEventAvailability({ webinarId })` currently uses `webinar.appointment?.slotsOfAppointment?.length` as `currentParticipants`. That is the number of master slot rows, not the number of registered users.

For a webinar with one appointment slot and many enrolled users, this path still reports `currentParticipants = 1`. `joinWaitlist()` relies on that helper, so a genuinely full webinar can be treated as if it still has capacity and users get blocked from joining the waitlist with “Spots are still available”.

**Proposed fix**
- Reuse the same participant-counting logic as checkout (`countWebinarParticipants`) instead of slot-row count.
- Keep consultant/host exclusions consistent with checkout so capacity policy is identical in both paths.
- Add a regression test for a webinar with one slot and multiple enrolled users where `joinWaitlist()` must return WAITLIST, not “spots available”.

## 2. P1: `NOTIFIED` waitlist users still do not actually hold a reserved seat

Files:
- `lib/waitlist/slot-handler.ts` (`205-216`)
- `lib/payments/operations/checkout.ts` (`1243-1250`, `1357-1362`, `1800-1803`)

`handleWaitlistResponse()` still says “redirect them to checkout with a reserved spot”, but nothing in checkout capacity validation subtracts active `NOTIFIED` entries or otherwise honors that reservation. The `fromWaitlist` flag is only used after successful payment to mark the waitlist entry `BOOKED`.

That means a user can be told they have a reserved seat, spend time going through checkout, and then lose the seat to a normal direct checkout because the notified hold was never represented in capacity.

**Proposed fix**
- Either treat `NOTIFIED` entries as temporary reserved capacity during the response window, or stop presenting them as reserved and handle them as first-come-first-served.
- If reservation is intended, capacity checks need to subtract active `NOTIFIED` holds and `fromWaitlist` checkout must validate against the matching waitlist row.
- Add a race test: notify waitlist user A, then attempt normal checkout by user B before A pays.

## 3. P1: one-on-one participant removal still corrupts consultation/subscription bookings

Files:
- `app/api/participants/consultations/[consultationId]/route.ts` (`146-160`)
- `app/api/participants/subscriptions/[subscriptionId]/route.ts` (`141-155`)
- `app/dashboard/consultant/[consultantId]/(features)/appointments/participants/consultations/[consultationId]/page.tsx` (`138-145`)
- `app/dashboard/consultant/[consultantId]/(features)/appointments/participants/subscriptions/[subscriptionId]/page.tsx` (`138-145`)

The consultant UI still renders a generic `Remove` action for every participant in 1:1 consultations and subscriptions, and the API still handles it by blindly disconnecting that user from every slot.

That leaves the appointment and parent booking intact while one side has been removed from the slot-user M2M. The system can therefore keep a booked consultation/subscription whose meeting slots only contain the consultant or only contain the consultee, which breaks access, notifications, and any logic that assumes 1:1 bookings always have both parties attached.

**Proposed fix**
- Do not expose participant-removal for 1:1 bookings, or convert it into a structured cancel/reschedule action instead of raw M2M disconnects.
- Add a server-side guard that refuses to disconnect participants from consultation/subscription slots directly.
- Add a regression test asserting a 1:1 booking cannot end up with fewer than two attached users while still remaining active.

## 4. P2: class consultee event API still filters included waitlist rows with the wrong ID type

File:
- `app/api/events/classes/route.ts` (`143-146`)

The `where` clause correctly finds classes where the current consultee is in the waitlist by traversing `user.consulteeProfile.id`, but the included `waitlist` relation is still filtered with `userId: consulteeProfileId`.

`Waitlist.userId` stores the `User.id`, not the `ConsulteeProfile.id`, so the query returns the class row but strips out the matching waitlist entry from the payload. That makes downstream UI lose per-user waitlist metadata such as status/position.

**Proposed fix**
- Filter the included waitlist rows using the same `user.consulteeProfile.id` path as the outer query, or resolve the real `userId` first and use that consistently.
- Add a test that a waitlisted class includes the current user’s waitlist row in the consultee API response.

## 5. P2: class `skipPayment` waitlist insertion still claims success from inside a transaction that immediately rolls back

File:
- `lib/payments/operations/checkout.ts` (`1362-1372`)
- Compare with the webinar out-of-transaction fallback at `1843-1919`

When a class is full and `skipPayment` is true, `handleClassCheckout()` creates a waitlist row via `tx.waitlist.create(...)` and then throws `"Class is full. Added to waitlist."`.

Because the throw aborts the surrounding transaction, that waitlist insert is rolled back. Unlike webinars, there is no matching outer catch path that recreates the class waitlist row outside the transaction. So the caller can receive an “added to waitlist” message while no waitlist record actually exists.

**Proposed fix**
- Move class waitlist creation to the same out-of-transaction fallback pattern already used for webinars.
- Or stop using exceptions for control flow here and return a structured “waitlisted” result that commits cleanly.
- Add a regression test that a full class with `skipPayment=true` leaves behind a persisted waitlist row.

## Checklist additions worth adding

- [ ] Webinar waitlist availability counts enrolled users, not slot rows.
- [ ] `NOTIFIED` waitlist offers either reserve capacity explicitly or the UI copy avoids promising a reserved spot.
- [ ] 1:1 consultation/subscription flows never allow raw participant-disconnects that leave active bookings with one user attached.
- [ ] Consultee webinar/class APIs use consistent `User.id` vs `ConsulteeProfile.id` scoping for included waitlist rows.
- [ ] Waitlist creation paths never report success from inside a transaction that is immediately rolled back.

teetangh:
I re-verified the latest fixes and then pushed deeper into free trials, waitlists, participant removal, and cancellation flows. The next cluster I found is below.

1. **P1: `/api/trials/[trialId]` still lets any participant drive consultant-only state transitions**  
   Files: `app/api/trials/[trialId]/route.ts`  
   The `PATCH` handler only checks that the caller is *related* to the trial (`consulteeProfileId` or `consultantProfileId`) and then applies the generic transition table. There is no role gate around the actual transition being requested. In the current code, a consultee can take a `PENDING` trial straight to `SCHEDULED` by sending `slotData`, can mark a scheduled trial `COMPLETED`, and can exercise consultant-only control paths that the UI only exposes to the consultant. This is a direct state-integrity problem, not just a UI gap. The route should enforce role-based transitions explicitly, e.g. consultant-only for `SCHEDULED` / `REJECTED`, both parties or policy-specific handling for `CANCELLED`, and system/consultant-only for `COMPLETED`.

2. **P1: Trial scheduling still uses stale overlap logic instead of the shared occupancy policy**  
   Files: `app/api/trials/[trialId]/route.ts`, `utils/slotAllocation/occupancyPolicy.ts`  
   `validateSlotAvailability()` still hard-codes consultations/subscriptions as occupied only when `requestStatus === APPROVED`, webinars/classes only when `status === SCHEDULED`, and trials only when `status === SCHEDULED`. The hardened booking logic elsewhere now treats `PENDING`, `APPROVED_PENDING_PAYMENT`, and `IN_PROGRESS` as occupied too. That means a trial can still be scheduled on top of a pending consultation hold, an approved-pending-payment subscription slot, or an in-progress webinar/class. This route should reuse `buildOccupiedAppointmentFilter()` (or the same canonical status set) instead of maintaining a divergent hand-rolled copy.

3. **P1: Trial creation still trusts mismatched `consultantProfileId` and `subscriptionPlanId`**  
   Files: `app/api/trials/route.ts`  
   The `POST /api/trials` handler verifies that the plan exists and that free trial is enabled, but it never checks that `subscriptionPlan.consultantProfileId === consultantProfileId` from the request body. A caller can therefore submit consultant A's `consultantProfileId` with consultant B's `subscriptionPlanId`, producing a trial row owned by consultant A but titled/logged/notified from consultant B's plan. That corrupts activity logs, notifications, and the consultant-side trial queue. The fix is to derive `consultantProfileId` from the plan server-side or at minimum reject mismatches.

4. **P2: Trial eligibility is still queryable without an ownership check and leaks prior trial metadata**  
   Files: `app/api/trials/check-eligibility/route.ts`  
   I do not see any session/auth check in this handler. It accepts arbitrary `consulteeProfileId` and `consultantProfileId`, then returns whether a trial exists plus `status`, `subscriptionPlanId`, and `requestedAt`. Even if the intended UI is authenticated, the API itself is still trusting caller-supplied profile IDs. At minimum this should require auth and enforce that the caller owns `consulteeProfileId` unless privileged.

5. **P1: The scheduled-trial cancellation path used by the consultant UI still only flips status to `CANCELLED`**  
   Files: `app/dashboard/consultant/[consultantId]/(features)/trials/TrialsTab.tsx`, `app/api/trials/[trialId]/route.ts`, `app/api/meetings/[meetingId]/validate-access/route.ts`  
   The consultant dashboard still cancels trials via `DELETE /api/trials/[trialId]`. That delete path only updates the trial status; it does not clean up the linked appointment/slots, does not clear any meeting artifact, and does not send the cancellation notification that the `PATCH` cancellation branch sends. Because meeting access is still granted from `slotOfAppointment.user` / appointment ownership, a cancelled trial can keep a live joinable meeting path even after the trial is marked cancelled. This needs parity with the real cancellation path: clean up or invalidate the appointment/meeting resources and notify both parties.

6. **P1: One-on-one participant removal is still breaking active consultations and subscriptions**  
   Files: `app/api/participants/consultations/[consultationId]/route.ts`, `app/api/participants/subscriptions/[subscriptionId]/route.ts`, `app/dashboard/consultant/[consultantId]/(features)/appointments/participants/consultations/[consultationId]/page.tsx`, `app/dashboard/consultant/[consultantId]/(features)/appointments/participants/subscriptions/[subscriptionId]/page.tsx`  
   The consultant participant pages still render a generic `Remove` action for consultations and subscriptions, and the API still responds by disconnecting the selected user from slot-user relations while leaving the booking itself active. That produces a half-detached 1:1 booking where the appointment, payment, and meeting assumptions remain active but only one side is still attached to the actual slot rows. For one-on-one products this action should not exist as a raw participant disconnect. It should be replaced with cancel/reschedule semantics, and the server should reject direct participant removal for these event types.

7. **P1: Waitlist-only webinars and classes are still missing from the consultee appointments feed**  
   Files: `app/api/dashboard/consultee/[consulteeId]/events/route.ts`  
   The comments say the webinar/class queries include waitlist entries, but the `where` clauses still only match slot membership for both event types. The `include.waitlist` block is present, but users who are only `WAITING` / `NOTIFIED` and not yet attached to appointment slots still do not get those events in the dashboard payload. This is still why waitlist-only webinars/classes disappear from the consultee appointments view.

8. **P1: Whole-event webinar/class cancellation still notifies the waitlist for an event that was just cancelled**  
   Files: `app/api/appointments/[appointmentId]/cancel/route.ts`  
   The cancel route still unconditionally calls `handleSlotOpening()` whenever a webinar or class appointment is cancelled. For whole-event cancellation, there is no real spot opening; the event itself has just been moved to `CANCELLED`. That means waitlist users can still be notified about a seat on an event that no longer exists. The route needs to distinguish between participant/seat release versus full event cancellation and only notify the waitlist in the former case.

Checklist additions I would fold into `Docs of Booking Algorithm/checklist.md`:
- Free-trial routes must enforce role-based state transitions, not just participant membership.
- Any trial/availability conflict check must reuse the canonical occupancy policy.
- Trial creation must derive consultant ownership from the selected subscription plan.
- Trial eligibility checks must be authenticated and owner-scoped.
- Trial cancellation paths must clean up linked appointment/meeting artifacts and send parity notifications.
- One-on-one products must never allow raw participant disconnects.
- Consultee event feeds must include waitlist-only webinar/class entries.
- Waitlist notifications must never fire after full event cancellation.

teetangh:
I pushed the next review pass into the collaborator surface. This batch is more about permission symmetry and private-schedule correctness than the core slot algorithm, but these are still real product bugs because the collaborator UI now treats collaborated events as first-class.

1. **P1: Collaborated webinars/classes are surfaced in the planner, but meeting access still only recognizes the owning consultant**  
   Files: `app/api/dashboard/consultant/[consultantId]/planner/route.ts`, `app/dashboard/consultant/[consultantId]/(features)/planner/components/EventManagementDashboard.tsx`, `app/api/meetings/[meetingId]/validate-access/route.ts`  
   The planner explicitly includes accepted collaborated webinars/classes (`collabWebinarsRaw` / `collabClassesRaw`) and lets the collaborator press `Join Meeting`. But `validate-access` still computes host access only from `appointment.webinar.webinarPlan.consultantProfileId` / `appointment.class.classPlan.consultantProfileId`, i.e. the plan owner only. Accepted collaborators are neither treated as hosts nor guaranteed to be attached to slot-user rows, so they can be shown a joinable collaborated event, create/open its meeting, and then get rejected at access validation. The access layer needs to recognize accepted collaborators on webinar/class plans as authorized hosts or moderators.

2. **P1: Recording APIs still deny accepted collaborators even though collaborated events are part of the consultant workflow**  
   Files: `app/api/plans/webinars/[webinarPlanId]/recordings/route.ts`, `app/api/plans/classes/[classPlanId]/recordings/route.ts`, `app/api/stream/recordings/[recordingId]/route.ts`  
   All three recording routes still grant consultant access only when `session.user.consultantProfileId === plan.consultantProfileId` (or the equivalent owner lookup through the appointment). That excludes accepted collaborators entirely. Given that the PR now exposes collaborated events in planner/collaboration surfaces, collaborators can participate operationally but still cannot view plan recordings or recording details for those same events. The access checks should include accepted webinar/class collaborators, not only the owner consultant.

3. **P1: Shared webinar/class event authorization is still owner-only, so collaborator-facing event APIs remain blocked**  
   Files: `lib/auth-helpers.ts`, `app/api/events/webinars/[webinarId]/route.ts`, `app/api/events/classes/[classId]/route.ts`, `app/api/events/webinars/[webinarId]/allocate/route.ts`, `app/api/events/classes/[classId]/allocate/route.ts`, `app/api/events/webinars/[webinarId]/validate/route.ts`, `app/api/events/classes/[classId]/validate/route.ts`  
   `authorizeEventAccess()` still authorizes webinar/class access solely by comparing the caller’s consultant profile to the plan owner profile. The helper is then reused by the webinar/class detail, validate, and allocate routes. So even after the planner route surfaces collaborated events, an accepted collaborator still cannot hit the underlying event APIs that back inspection/allocation/validation flows. This is the same owner-only assumption as the meeting/recording gap, just in the shared route layer. Either collaborators are operational actors for these event types, in which case these routes need collaborator-aware auth, or the collaborator-facing planner/actions need to stop pretending otherwise.

4. **P1: Collaborator availability API leaks a consultant’s private schedule and raw booking windows to any authenticated user**  
   Files: `app/api/collaborators/[consultantProfileId]/availability/route.ts`  
   This route only checks that the caller is authenticated. After that, any logged-in user can request any consultant profile ID and receive weekly availability, custom availability, and concrete booked-slot intervals for that consultant. That is much more sensitive than the public availability APIs and should not be exposed on a generic authenticated basis. The route should verify a legitimate relationship (owner, accepted collaborator on a shared plan, admin/staff, etc.) before returning private scheduling data.

5. **P1: The collaborator availability overlay still ignores collaborated commitments and still uses non-canonical day filters**  
   Files: `app/api/collaborators/[consultantProfileId]/availability/route.ts`  
   Even aside from the auth leak, the endpoint still only loads booked webinar/class slots where the consultant is the plan owner (`webinarPlan.consultantProfileId` / `classPlan.consultantProfileId`). Accepted collaborator commitments on someone else’s webinar/class are omitted, so the overlay can show a collaborator as free when they are already committed to a shared event. The date filters are also still the older `startsAt within day` / `custom.startsAt within day` style rather than canonical overlap semantics, so overnight or spanning slots can still drop out of the day view. This route should use the same occupancy/overlap policy as the rest of the booking system and include collaborated webinar/class obligations.

6. **P2: Revenue-split preview endpoints still expose collaborator payout structure to any authenticated user**  
   Files: `app/api/collaborations/webinar/[planId]/revenue-split/route.ts`, `app/api/collaborations/class/[planId]/revenue-split/route.ts`  
   These endpoints currently require only a logged-in session, then return the split for any plan ID and amount. That leaks collaborator consultant profile IDs, effective shares, and owner/collaborator payout structure for arbitrary plans. Even if today this is only used by internal UI, the API itself is under-scoped. It should be limited to the plan owner, accepted collaborators on that plan, or privileged roles.

Checklist additions I’d make from this pass:
- Any collaborator-visible event must have collaborator-consistent auth across planner, meeting join, recordings, detail, validate, and allocate routes.
- Private collaborator availability endpoints must be relationship-scoped, not merely authenticated.
- Collaborator availability overlays must include collaborated event commitments and use canonical overlap semantics.
- Revenue-split preview endpoints must be plan-scoped to authorized collaborators/owners only.

shubham79a (approved):

PR #504, opened 2026-03-20
fix: resolve booking modal wrong times — local minutes stored as UTC

teetangh:
## Summary

Shubham set a **Monday 1:00 PM – 3:00 PM IST** weekly slot during onboarding. The booking modal displayed **6:30 PM – 8:30 PM** — shifted by exactly +5h30m (the IST offset). The profile grid appeared correct only by coincidence (it was reading the raw local minutes as-if they were UTC).

### Root cause

The onboarding form converted local HH:MM → UTC minutes inside a **`useEffect`**, but form submission ran **synchronously** via `handleSubmit(onSubmitForm)`. If the user clicked "Next" before the async effect fired, stale/unconverted local values were submitted as UTC. The settings path (`SettingsTab.tsx`) didn't have this bug because it calls `formatSlotsForApi()` synchronously inside the submit handler.

### What changed

| File | Change |
|------|--------|
| `ConsultantPreferredScheduleForm.tsx` | Removed the two `useEffect`s that called `buildWeeklySlotsForSave`/`buildCustomSlotsForSave`. Conversion now happens synchronously in `onSubmitForm`, matching the working SettingsTab pattern. |
| `utils/dateTimeUtils.ts` | `convertTimezoneToUtc` now uses explicit `new Date(year, month-1, day, hours, minutes)` constructor instead of parsing `"1970-01-01T13:00:00"` strings — avoids browser ambiguity where TZ-less ISO strings may be parsed as UTC, making `fromZonedTime` a no-op. |
| `utils/schedule/formatting.ts` | `buildWeeklySlotsForSave` now computes the actual UTC day-of-week after timezone conversion (e.g., IST Monday 1:00 AM → UTC Sunday 7:30 PM shifts the day back). Also detects when a same-day local slot becomes overnight in UTC. Exported `shiftDayOfWeek` for reuse. |
| `utils/onboarding-server.ts` | Added `console.info` logging before `createMany` to aid future debugging of slot values hitting the DB. |

### DB fix (already applied)

Shubham's existing slot was corrected via Supabase SQL:
```
startTimeUtc: 780 → 450  (7:30 AM UTC = 1:00 PM IST)
endTimeUtc:   900 → 570  (9:30 AM UTC = 3:00 PM IST)
```

Audit of all consultants with `utcOffsetMinutes != 0` confirmed only Shubham's production data was affected — all other rows are E2E test seed data with intentionally set values.

## Test plan

- [ ] Verify Shubham's profile: booking modal should show 1:00–2:00 / 1:30–2:30 / 2:00–3:00 PM IST slots (not 6:30 PM+)
- [ ] Create a new test consultant via onboarding with an IST weekly slot → check DB `startTimeUtc`/`endTimeUtc` are actual UTC values (local minus 330)
- [ ] Modify a weekly slot via Settings → verify DB values remain correct (regression check — settings path was already working)
- [ ] View a consultant profile from a UTC browser → verify displayed times convert correctly
- [ ] Test overnight slot edge case: set a slot crossing midnight (e.g., 11 PM – 1 AM IST) → verify `startDay` and `endDay` are correct in DB

### DIFF
--- a/app/form/onboarding/components/ConsultantPreferredScheduleForm.tsx
+++ b/app/form/onboarding/components/ConsultantPreferredScheduleForm.tsx
@@ -134,17 +134,10 @@ const ConsultantPreferredScheduleForm: React.FC<Props> = ({
     }
   }, [initialData, timezone, timezoneLoading]);
 
-  // Format weekly slots for API
-  useEffect(() => {
-    if (!timezone) return;
-    setValue("weeklySlots", buildWeeklySlotsForSave(weeklySlots, timezone));
-  }, [weeklySlots, setValue, timezone]);
-
-  // Format custom slots for API
-  useEffect(() => {
-    if (!timezone) return;
-    setValue("customSlots", buildCustomSlotsForSave(customSlots, timezone));
-  }, [customSlots, setValue, timezone]);
+  // NOTE: Conversion from local HH:MM → UTC minutes is done synchronously
+  // in onSubmitForm (not in a useEffect) to avoid a race condition where
+  // the form could submit stale/unconverted values if the user clicks "Next"
+  // before the useEffect fires. This mirrors the working pattern in SettingsTab.
 
   const handleAddSlot = useCallback(
     (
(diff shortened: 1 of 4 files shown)


teetangh:
Wrong diagnosis — the save path was correct. Real bug is format() ignoring timeZone on UTC servers. New PR incoming with formatInTimeZone fix.

Issue #248, opened 2025-12-14
🐌 Performance: Stream Chat sync runs on every dashboard load causing excessive API calls


teetangh:
## Summary

Every time a user opens their dashboard, the Stream Chat initialization triggers a full channel sync that makes **5-10 API calls per event** (webinar/class), causing:
- Slow dashboard load times (15-20+ seconds observed)
- Excessive logging spam in the console
- Unnecessary API quota usage on Stream
- Poor user experience

## Root Cause

The `syncUserEventChannels` function in `StreamProvider.tsx` runs during `connectChat()` on every page load, even when the user is already a member of all their channels.

### Primary Files
- `providers/StreamProvider.tsx` (lines 159-174) - Triggers sync on every chat connection
- `actions/stream/chat/event-channel.action.ts` - Contains the inefficient sync logic

### Flow
1. `StreamProvider.tsx` → `connectChat()` (line 128)
2. → `syncUserEventChannels(userId)` (line 165)
3. → For each webinar/class, calls `addUserToEventChannel()`
4. → Each call makes multiple API requests:
   - `checkEventChannelExists()` - queries Stream API
   - Database query for event details
   - `channel.update()` - updates name (even if unchanged)
   - `upsertUserToStream()` - upserts user (redundant)
   - `channel.addMembers()` - adds member (even if already member)

## Current Behavior (Problem)

```
// Example console output on dashboard load:
Channel webinar-cmj3dled403klkbhnb8bddci1 exists: true {...huge object...}
Updating channel webinar-cmj3dled403klkbhnb8bddci1 name to "Advanced Topics Webinar"
Upserting user cmj3d3bmf0003kbhnqhob8pnj with role admin
Channel webinar-cmj3dled403klkbhnb8bddci1 exists. Adding member cmj3d3bmf0003kbhnqhob8pnj...
Added user cmj3d3bmf0003kbhnqhob8pnj to existing webinar channel...
// Repeats for EVERY webinar and class (10+ times in this case)
```

## Impact

| Metric | Current | Expected |
|--------|---------|----------|
| API calls per dashboard load | 50-100+ | 1-2 |
| Load time | 15-20s | <2s |
| Console log lines | 200+ | 5-10 |
| Stream API quota impact | High | Minimal |

### Option 1: Event-Driven Sync (Recommended)
Add users to channels **when they subscribe/join**, not on every login.

```typescript
// In subscription/enrollment handlers:
[...]

PR #532, opened 2026-03-28
fix: waitlist cron contention + Stream Chat sync perf (#514, #515, #516, #248)

teetangh:
## Summary

- **#514, #515, #516** — Three waitlist job failures all on Mar 22 within 1 hour. Root cause: 6 cron jobs fire at `0 * * * *` simultaneously, exhausting the DB connection pool. Also, every failure creates a new GitHub issue (no deduplication).
- **#248** — Stream Chat `syncUserEventChannels` runs on every dashboard page load, making 50-100+ Stream API calls. Root cause: the in-memory `initialSyncCompletedUsers` Set resets on every page load because client-side JS modules re-evaluate from scratch on navigation/refresh.

### Waitlist cron jobs
- Stagger `process-waitlist-expirations` from `:00` to `:05` to avoid DB pool contention with 5 other hourly jobs
- Add `DIRECT_URL` env var to both waitlist workflows (fallback if `DATABASE_URL` is unavailable)
- Deduplicate failure notifications: search for existing open issue before creating a new one; append a comment to the existing issue instead

### Stream Chat sync
- Replace in-memory `initialSyncCompletedUsers` guard with `sessionStorage` — persists across page reloads within the same tab session
- Clear sync flags on sign-out (`disconnectStreamClients`) so re-login triggers a fresh sync
- Fix `useEffect` dependency instability: use refs for `connectServices`/`disconnect` so the effect only re-fires on stable values (`userDetails.id`, `isLoading`, `apiKey`)
- **Net effect: 50-100 Stream API calls per dashboard load → 0 on reload**

## Test plan

- [ ] `npx tsc --noEmit` passes (verified)
- [ ] `npm run test` — 676/676 pass (verified)
- [ ] Stream sync: open dashboard, refresh page — no `syncUserEventChannels` calls in Network tab on second load
- [ ] Stream sync: sign out + sign back in — sync should run once on first load after re-login
- [ ] Waitlist: trigger `workflow_dispatch` on both workflows — should succeed with staggered timing
- [ ] Waitlist: if a job fails twice, only 1 GitHub issue should exist (second failure appends a comment)

### DIFF
--- a/.github/workflows/process-waitlist-expirations.yml
+++ b/.github/workflows/process-waitlist-expirations.yml
@@ -44,22 +44,46 @@ jobs:
         run: npx prisma generate
 
       - name: Run expiration processor
-        run: npx tsx scripts/waitlist/process-expired-notifications.ts
+        id: run_script
+        run: npx tsx scripts/waitlist/process-expired-notifications.ts 2>&1 | tee /tmp/job-output.txt
         env:
           DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}
+        continue-on-error: true
+
+      - name: Capture error output
+        if: steps.run_script.outcome == 'failure'
+        id: error_output
+        run: |
+          ERROR_LOG=$(tail -50 /tmp/job-output.txt || echo "No output captured")
+          echo "log<<EOFLOG" >> $GITHUB_OUTPUT
+          echo "$ERROR_LOG" >> $GITHUB_OUTPUT
+          echo "EOFLOG" >> $GITHUB_OUTPUT
 
       - name: Summary
-        if: success()
+        if: steps.run_script.outcome == 'success'
         run: |
           echo "## Waitlist Expiration Processing Complete" >> $GITHUB_STEP_SUMMARY
           echo "Processed expired notifications and notified next users in queue." >> $GITHUB_STEP_SUMMARY
 
       - name: Notify on failure
-        if: failure()
+        if: steps.run_script.outcome == 'failure'
         uses: actions/github-script@v7
         with:
           script: |
             const title = 'Waitlist Expiration Job Failed';
+            const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+            const errorLog = process.env.ERROR_LOG || 'No error output captured';
+            const body = [
+              `The waitlist expiration processing job failed at ${new Date().toISOString()}`,
+              '',
+              '**Error output (last 50 lines):**',
+              '```',
+              errorLog,
+              '```',
+              '',
+              `See full run: ${runUrl}`,
+            ].join('\n');
+
             const { data: existing } = await github.rest.issues.listForRepo({
               owner: context.repo.owner,
               repo: context.repo.repo,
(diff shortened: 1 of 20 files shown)


teetangh:
I reviewed PR #532 locally against `dev`, with extra attention on the Stream changes because that is where most of the behavioral risk is concentrated.

Overall direction makes sense: reducing N+1 Stream sync work, staggering the waitlist cron, and hardening client connection lifecycle are all worthwhile. The concerns below are the places where I think the current implementation still has correctness or regression risk.

1. `disconnectStreamClients()` only disconnects chat, not video.

In `providers/StreamProvider.tsx`, the function comment says this is meant to "Disconnect all Stream clients and clear global state", but the implementation only waits on `globalChatClient.disconnectUser()` and then nulls `globalVideoClient` directly.

Why this matters:
- Nulling the JS reference is not the same thing as disconnecting the authenticated Stream Video client.
- The installed Stream Video SDK does expose `disconnectUser()`, so the current code is leaving the previous video session/socket lifecycle to chance instead of explicitly closing it.
- That can show up as stale authenticated state after logout/account switch, or as reconnect weirdness later when a new user session is established.

Concretely, the risk is that the PR improves chat teardown semantics but leaves video teardown incomplete, so the two halves of the provider no longer behave symmetrically during logout.

2. `searchUsersWithRelationships()` changed subscription semantics in a subtle but user-visible way.

The new batched query approach in `actions/stream/chat/user.action.ts` is a good optimization in principle, but both batched subscription queries dropped the `schedulingPeriodEndsAt >= now` constraint.

Why this matters:
- The older per-user helper still treats only active subscriptions as a valid relationship.
- The new batched path will mark expired consultant/consultee pairs as related as long as the subscription row is still `APPROVED` or `SCHEDULED`.
- That means the DM search results can now surface people as connected when the subscription access window has already ended.

This is the kind of regression that is easy to miss in manual testing because the feature still "works", but the relationship model is no longer consistent with the existing helper logic.

3. The current refactor breaks the Stream user-action tests.

I ran the targeted Stream suite locally with:

`npm test -- --runInBand __tests__/stream/user-actions.test.ts __tests__/stream/event-channel-actions.test.ts __tests__/stream/channel-actions.test.ts`

Result:
- `event-channel-actions.test.ts` passed
- `channel-actions.test.ts` passed
- `user-actions.test.ts` failed

The failure is a runtime `TypeError: Cannot read properties of undefined (reading 'then')` inside the new batched relationship logic.

Why this matters:
- The refactor replaced the previous per-user `findFirst()` pattern with batched `findMany()` calls.
- The tests and shared mocks were not updated to match the new call shape.
- That means the exact area this PR changes most heavily is currently not protected by CI in its present form.

Even if the production code path is directionally correct, merging while the targeted unit coverage is red makes this harder to trust and easier to regress further.

Netting it out:
- I do not think the waitlist workflow changes are the risky part of this PR.
- The two places I would treat as blocking are the incomplete Stream Video disconnect path and the subscription-relationship semantic drift in the batched user search.
- The failing test coverage on the refactored search path compounds that risk because it removes the safety net exactly where the behavior changed.

If this PR is going forward as-is, I would want those Stream issues resolved first or explicitly carved out into a follow-up with clear acknowledgement that the current branch is shipping behavior changes in logout/session teardown and DM relationship matching.

PR #632, opened 2026-04-05
fix(consultant-dashboard): production-grade join button with proper time-window logic

teetangh:
## Summary
- **Bug:** Join button only enabled during a 5-minute window before session start, disabled once session was in progress
- **Fix:** Created `joinState.ts` utility mirroring the consultee dashboard's proven 15-min join window pattern, decoupled from display-status string comparison
- **Cleanup:** Eliminated dual field naming (`slotStartTimeInUTC` → `startsAt`) across 7 consumer files, added `meetingSession` to `TAppointment` type

## Changes
| Area | What |
|------|------|
| **New utility** | `joinState.ts` — `getSlotJoinState()` + `getJoinableSlot()` with 15-min pre-join window |
| **API** | Consultant dashboard API stops renaming `startsAt`/`endsAt`, adds `meetingSession` include |
| **Types** | `TAppointment` slots now include `meetingSession { id, endedAt }` |
| **Components** | AppointmentsTab, HomeTab, PaginatedAppointments use `getJoinableSlot()` instead of string matching |
| **Display** | New "In Progress" (emerald) and "Starting soon" (amber) badge statuses |
| **Dev mode** | All join buttons always enabled with "Join (Dev)" label |

## Join Window Logic (Production)
```
──────────────────────────────────
  -15min    startsAt        endsAt
    │          │               │
    ├──────────┼───────────────┤
    │  JOINABLE (enabled)      │
────┘                          └────
 NOT AVAILABLE           NOT AVAILABLE
```

## Test plan
- [ ] TypeScript compiles clean (`tsc --noEmit` passes)
- [ ] Dev mode: all appointments show "Join (Dev)" button, always clickable
- [ ] Prod: button enables 15 min before start, stays enabled through session end
- [ ] Tentative slots: join button disabled
- [ ] Cancelled/Rescheduled slots: excluded from join logic
- [ ] Multi-session subscriptions: only current/next session joinable
- [ ] Planner EventCard unchanged (already has correct pattern)
- [ ] "In Progress" and "Starting soon" badges render with correct colors

### DIFF
--- a/app/api/dashboard/consultant/[consultantId]/route.ts
+++ b/app/api/dashboard/consultant/[consultantId]/route.ts
@@ -18,6 +18,7 @@ const userSelectFields = {
 
 const appointmentInclude = {
   slotsOfAppointment: {
+    orderBy: { startsAt: "asc" as const },
     include: {
       user: {
         select: userSelectFields,
--- a/app/api/dashboard/consultant/[consultantId]/route.ts
+++ b/app/api/dashboard/consultant/[consultantId]/route.ts
@@ -22,6 +22,9 @@ const appointmentInclude = {
       user: {
         select: userSelectFields,
       },
+      meetingSession: {
+        select: { id: true, endedAt: true },
+      },
     },
   },
   consultation: {
@@ -435,9 +438,11 @@ export async function GET(
         appointmentType: appointment.appointmentType,
         slotsOfAppointment: appointment.slotsOfAppointment.map((slot) => ({
           id: slot.id,
-          slotStartTimeInUTC: new Date(slot.startsAt),
-          slotEndTimeInUTC: slot.endsAt ? new Date(slot.endsAt) : null,
+          startsAt: slot.startsAt,
+          endsAt: slot.endsAt,
           isTentative: slot.isTentative,
+          completionStatus: slot.completionStatus,
+          meetingSession: slot.meetingSession ?? null,
           user: Array.isArray(slot.user)
             ? slot.user.map((u) => ({
                 id: u.id,
--- a/app/api/slots/appointments/route.ts
+++ b/app/api/slots/appointments/route.ts
@@ -336,6 +336,7 @@ async function getAppointments(
     where: whereClause,
     include: {
       slotsOfAppointment: {
+        orderBy: { startsAt: "asc" },
         include: {
           user: true,
           meetingSession: {
--- a/app/api/slots/appointments/route.ts
+++ b/app/api/slots/appointments/route.ts
@@ -338,6 +338,9 @@ async function getAppointments(
       slotsOfAppointment: {
         include: {
           user: true,
+          meetingSession: {
+            select: { id: true, endedAt: true },
+          },
         },
       },
       consultation: {
--- a/app/dashboard/consultant/[consultantId]/(features)/appointments/AppointmentsTab.tsx
+++ b/app/dashboard/consultant/[consultantId]/(features)/appointments/AppointmentsTab.tsx
@@ -64,6 +64,7 @@ import {
   getParticipantManagementUrl,
   supportsParticipantManagement,
 } from "./utils/participantHelpers";
+import { getJoinableSlot } from "../../utils/joinState";
 
 export function AppointmentsTab({
   appointments,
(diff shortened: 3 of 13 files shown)


teetangh:
/btwby the way, can the consultee join 10 minutes before, during the 10 minutes, and also
      anytime during the session? Also, are we handling it across multiple time zones and
      several slots and sessions and daylight saving time, etc.? Have you taken everything into
      account? Think of this as a production-grade app and think of all possible edge cases,
      corner cases, and stuff.

    SessionTimeline.tsx, HomeTab.tsx).

    Consultant Join Window (our fix)

    We're using 15 minutes before through endsAt for the consultant side. This is intentionally
    wider than the consultee's 10 minutes because the consultant (host) typically needs to set
    up before participants arrive. This is a common pattern in meeting platforms.

    Timezone Handling

    Yes, this is timezone-safe. Here's why:

    - All slot times are stored as UTC timestamps with timezone offset in Postgres
    (timestamptz): e.g., 2026-04-05 12:30:00+00
    - JavaScript's new Date("2026-04-05T12:30:00Z") converts this to an internal millisecond
    epoch (UTC-based)
    - All comparisons (start <= now && now <= end) operate on these epoch milliseconds —
    timezone-agnostic
    - A consultant in IST (UTC+5:30) and a consultee in PST (UTC-8) will both see the join
    button activate at the same absolute moment in time
    - The display formatting (e.g., "6:00 PM") uses Intl.DateTimeFormat / date-fns which
    respects the browser's local timezone — so each user sees their local time

    Daylight Saving Time (DST)

    Handled correctly because:
    - We store absolute UTC timestamps, not relative offsets like "Monday 10:00 AM local"
    - new Date() in the browser automatically accounts for the current DST offset
    - The comparison Date.now() >= joinWindow uses epoch milliseconds — DST doesn't affect this
    - The only DST risk would be in the slot creation phase (if a consultant creates a slot for
    "3:00 PM" and DST shifts between creation and session) — but since startsAt/endsAt are
    stored as absolute UTC timestamps at creation time, they remain correct

    Multiple Slots / Sessions

    Handled. The getJoinableSlot() function iterates through ALL slots:

    - Consultation (2 consecutive 30-min slots): If slot 1 is active, it's joinable. When slot 1
     ends and slot 2 starts, slot 2 becomes joinable.
    - Subscription (e.g., 12 sessions across 3 months): Only the currently active or
    upcoming-within-15-min slot is joinable. Past sessions return "ended", future sessions
    return "countdown".
    - Class (many sessions, shared across participants): Same logic — only the currently
    relevant slot is joinable.
    - Webinar (single session): Simple case — one slot, joinable within the window.

    Edge Cases Covered

    1. Gap between consecutive slots (slot 1 ends 10:30, slot 2 starts 11:00, now is 10:45):
    Slot 1 → "ended", slot 2 → "joinable" (within 15-min window). Correct.
    2. Tentative slots: isTentative === true → "disabled". Never joinable. Correct.
    3. Cancelled/Rescheduled slots: completionStatus === "CANCELLED" || "RESCHEDULED" →
    "disabled". Correct.
    4. Missing endsAt: In Prisma schema, endsAt is a required field (DateTime @db.Timestamptz),
    so it's never null from the database. No fallback needed — this is why we deleted the
    fallback logic.
    5. Session that started but meeting hasn't been created yet: getSlotJoinState() returns
    "joinable" based purely on time window. The meeting gets created on-demand via
    getOrCreateAppointmentMeeting() when the user clicks Join — so this is fine.
    6. Meeting ended early (meetingSession.endedAt is set but slot endsAt hasn't passed): Our
    join state is based on slot times, not meeting session state. The consultant can rejoin if
    the slot window is still open. This is intentional — allows rejoining after disconnections.
    7. Leap seconds / clock skew: Using Date.now() from the browser. Minor skew (seconds) is
    irrelevant for a 15-minute window.

    One Thing NOT Covered (but acceptable)

    The consultant's 15-minute window and the consultee's 10-minute window are different. This
    is intentional (host joins early to set up), but it means a consultant could be in the
    meeting room 5 minutes before any consultee can join. This is standard behavior for meeting
    platforms like Zoom/Teams.

PR #655, opened 2026-04-10
Enterprise foundation — org CRUD, billing modes, SSO, dashboard

teetangh:
# Enterprise foundation — full subsystem

This PR ships the complete enterprise subsystem on top of the existing
B2C marketplace: orgs, RBAC, SSO, billing/payouts, programs, audit, plus
the cross-cutting integrations (outbound webhooks, SCIM 2.0, DPDP §12
erasure, data export, retention crons, security headers, maintenance
tier-1, BILLING_ADMIN dashboard).

**Production readiness:** **100/100** per the May 2026 closeout.
**CI:** 957/957 jest, `tsc --noEmit` clean, lint clean.
**Schema migrations:** 4 applied to production via Supabase MCP
(`pr655_enterprise_lockdown_schema`, `pr655_schema_finalization_pre_mvp`,
`pr655_maintenance_per_tenant_tier1`).
**Defer list:** [#744 — Enterprise v1 post-MVP roadmap](https://github.com/Practitionist/familiarise_web/issues/744).

---

### 1. Organization lifecycle

- `Organization` model with the full capability matrix:
  `canSponsor` / `canHost` / `INERT`, funding modes
  (PERSONAL / WALLET / INVOICE / LICENSE), GST + MSME + PAN
  compliance fields, hierarchy columns (`parentId / rootId / depth`,
  UI deferred to #744), 14 status + capability indexes.
- Wizard flow → POST `/api/organizations` creates an `Organization`
  + `BillingAccount` + `OrgWorkspaceProfile` + member-OWNER membership
  in one transaction.
- Org status state machine: `PENDING_VERIFICATION → ACTIVE →
  SUSPENDED → DEACTIVATED` enforced via `requireOrgAccess` + admin
  verification routes.
- Public org marketplace at `/explore/enterprise/organisations`.
- `canHost: true` hard-gated server-side behind `ENABLE_HOST_ORGS`
  (rejects with typed `HOST_ORGS_GATED 400`) so an org never ends up
  half-functional. Replaces the prior WIP banner — per reviewer
  feedback "WIP banners are not production gates".

### 2. Membership + RBAC

- `MemberRole` enum: `OWNER · MAINTAINER · BILLING_ADMIN · MANAGER ·
  EXPERT · LEARNER · SUPPORT`. **`BILLING_ADMIN` is the new finance-
  team role** between MAINTAINER (rank 80) and MANAGER (rank 60) at
  rank 70.
[...]

Issue #661, opened 2026-04-10
Enterprise remaining features — consolidated tracker


teetangh:
# Enterprise Remaining Features

Consolidated tracker for all deferred enterprise work from PR #655 (`feature/enterprise`). The PR ships the core enterprise foundation (schema, APIs, dashboard, ORG_ADMIN onboarding, 3 billing modes, SSO, checkout integration, refund routing). These items are follow-up work that builds on that foundation.

Replaces closed issues: #656, #657, #658, #659, #660

---

## 1. Inngest Monthly Invoice Generation Cron

**Priority**: High
**Complexity**: Low (the logic already exists)

The manual `POST /api/organizations/[orgId]/billing/generate-invoice` endpoint works end-to-end: filters SUCCEEDED + ORG_INVOICED payments, nets out refunds, creates an OrganizationInvoice. This task wraps it in an automated Inngest scheduled function.

### Implementation plan
- **New file**: `lib/inngest/functions/generate-org-invoices.ts`
- **Schedule**: Runs on the 1st of each month at 00:00 UTC
- **Logic**:
  1. Query all OrganizationProfiles where `billingMode = "INVOICED_MONTHLY"` and `status = "ACTIVE"`
  2. For each org, call the same aggregation logic from `generate-invoice/route.ts`:
     - Find unbilled SUCCEEDED ORG_INVOICED payments
     - Net out refunds per payment
     - Skip orgs with no billable charges
     - Create OrganizationInvoice with status SENT, dueDate = paymentTermsDays from now
  3. Fire-and-forget email to `org.billingEmail` with the invoice summary
- **Wire into**: `app/api/inngest/route.ts` (register the function)
- **Testing**: Manual trigger via Inngest dashboard or `POST /api/inngest/send`

### Files
- `lib/inngest/functions/generate-org-invoices.ts` — **New**
- `app/api/inngest/route.ts` — Register the function
- Extract shared logic from `app/api/organizations/[orgId]/billing/generate-invoice/route.ts` into `lib/payments/operations/org-invoicing.ts`

---

## 2. Post-Invoice Refund Credit-Note System

**Priority**: Medium
**Complexity**: Medium

When an ORG_INVOICED payment is refunded AFTER it was already rolled into a PAID invoice, the current code only unbills the payment (`billableToOrgInvoiceId = null`) but does NOT update the `OrganizationInvoice.amount` or `items`. The invoice totals drift from reality.

### Implementation plan
[...]

Issue #438, opened 2026-02-14
Invoice System: PDF Generation, Download & Email Delivery


teetangh:
## Current State

The invoice **data infrastructure** is built and working — but the **rendering and delivery** layer is completely missing.

### What Exists

| Component | Status | Details |
|-----------|--------|---------|
| Prisma `Invoice` model | Done | invoiceNumber, amount, tax, items (JSON), pdfUrl, HSN codes |
| Auto-creation on payment | Done | `handlePaymentSuccess()` → Phase 2 → `createInvoiceFromPayment()` |
| Invoice number format | Done | `FAM-YYYYMM-XXXXX` with collision retry |
| GST/tax calculation | Done | 18% GST, SAC codes (999293 consulting, 999294 education) |
| API routes | Done | `/api/invoices`, `/api/invoices/[id]`, staff/admin views |
| Line items with HSN | Done | Stored as JSON in `items` field |
| Consultee dashboard tab | Done | Invoices tab in Payments page (currently shows empty) |

### What's Missing

| Component | Status | Impact |
|-----------|--------|--------|
| PDF generation | Not built | No downloadable invoices |
| PDF storage | Not built | `pdfUrl` field exists but is never populated |
| Invoice view page | Not built | No way to view invoice details in browser |
| Email delivery | Not built | No invoice sent to consultee after payment |
| Consultant invoice/receipt | Not built | Consultants have no payout receipts |
| Invoice template/branding | Not built | No Familiarise-branded invoice design |
| PDF library | Not installed | No pdfkit, jsPDF, react-pdf, or puppeteer in package.json |

---

### Legal/Tax Compliance (India)

GST-compliant invoices must include:
- [ ] Supplier name, address, GSTIN (Familiarise)
- [ ] Recipient name and address
- [ ] Invoice number (sequential, unique — already done: `FAM-YYYYMM-XXXXX`)
- [ ] Date of issue
- [ ] HSN/SAC code for each line item (already stored)
- [ ] Taxable value, GST rate (18%), CGST/SGST or IGST breakdown
- [ ] Total amount in words
- [ ] Digital signature (optional for <₹50K invoices)

### User-Facing

- Consultees should be able to view and download invoices for every successful payment
- Invoice should be available within seconds of payment (not async/delayed)
- Should work for all 4 service types (Consultation, Subscription, Webinar, Class)
- Credits/discounts applied should be reflected on the invoice

---

[...]

PR #666, opened 2026-04-13
feat(enterprise): PROVIDER/HYBRID orgs, PREPAID_UNLIMITED billing, 3-way earnings split

teetangh:
## Summary

Stacked on PR #655 (`feature/enterprise`). Implements the complete PROVIDER and HYBRID organization lifecycle, adds a 4th billing mode (PREPAID_UNLIMITED), and wires the 3-way earnings split for consultant agencies.

- **PROVIDER orgs**: Consultant agencies that host multiple consultants with configurable 3-way revenue split (platform/org/consultant). Gated by `ENABLE_PROVIDER_ORGS` feature flag.
- **HYBRID orgs**: Simultaneously BUYER (sponsor learners) + PROVIDER (host consultants) with independent financial streams.
- **PREPAID_UNLIMITED**: 4th billing mode for flat-fee enterprise contracts. No per-session billing — sessions are free for learners within a contract period.
- **3-way earnings split**: `resolveOrgSplit()` in earnings-service.ts creates both `ConsultantEarnings` + `OrganizationEarnings` atomically. Supports internal consultants (`earningsRecipient=ORGANIZATION`), per-consultant rate overrides, and platform-only mode (100% platform rate).
- **Payout pipeline**: `org-payout-service.ts` with Redis-locked batch creation, gateway processing (Razorpay/Stripe), and AES-256-GCM payout account encryption.
- **Public pages**: `/org/[slug]` org profile + `/explore/companies` directory + org badge on consultant cards.

### Key changes (65 files, +7408/-1976)

| Area | Files | What changed |
|------|-------|-------------|
| Schema | `prisma/schema.prisma` | 2 new enums, 10+ new fields, `PREPAID_UNLIMITED` billing mode, contract dates |
| Earnings | `lib/payments/payouts/earnings-service.ts` | `resolveOrgSplit()`, 3-way split, refund reversal, zero guard, org earnings summary |
| Payouts | `lib/payments/payouts/org-payout-service.ts` (new) | Batch creation, gateway processing, audit logging |
| Encryption | `lib/payments/payouts/account-crypto.ts` (new) | AES-256-GCM for bank account numbers |
| Checkout | `lib/payments/operations/checkout.ts` | PREPAID_UNLIMITED: skip gateway, contract expiry guard, `ORG_PREPAID` payment method |
| Refunds | `app/api/payments/refunds/route.ts` | `ORG_PREPAID` case: no financial action |
| API routes | 5 new routes | Consultant apply, admin verify, org payouts process, public org profile |
| Dashboard | 6 modified pages | Payouts (real data), overview (PROVIDER cards), billing (PREPAID_UNLIMITED), wizard (kind selector) |
| Public | 2 new pages | `/org/[slug]`, `/explore/companies` |
| Types | 7 files | PREPAID_UNLIMITED in all billing mode types/schemas/validators |
| Seed | `15a-create-organizations.ts` | 2 PROVIDER orgs, 1 HYBRID org with consultants, earnings, payouts |
| Docs | 15 new docs | Full enterprise documentation rewrite |

### Design decisions (confirmed with user)

| # | Decision | Choice |
|---|----------|--------|
| 1 | Discovery | Both: /explore with org badge + /explore/companies |
| 2 | Multi-org | Consultants can belong to multiple PROVIDER orgs |
| 3 | Verification | Platform admin approval required for PROVIDER orgs |
| 4 | Dual roles | Single role per org; higher roles get implicit learner access |
| 5 | Payout cadence | Configurable per-org (WEEKLY/BI_WEEKLY/MONTHLY) |
| 6 | HYBRID billing | Independent BUYER + PROVIDER financial streams |
| 7 | Internal consultants | `earningsRecipient` flag: CONSULTANT or ORGANIZATION |
| 8 | PREPAID_UNLIMITED | 4th billing mode with contract start/end dates |
| 9 | Prepaid refunds | No financial action — just cancel session + reverse earnings |

### Billing mode permutation matrix

| Mode | Who Pays | Per-Session Cost | Earnings Created |
|------|----------|-----------------|------------------|
| TAG_ONLY | Learner (own card) | Full price | Standard 80/20 |
| SEAT_PACK | Org (credit pool) | Deducted from pool | Standard 80/20 |
| INVOICED_MONTHLY | Org (monthly invoice) | Added to invoice | Standard 80/20 |
| PREPAID_UNLIMITED | Org (flat fee upfront) | Free (₹0) | Depends on earnings mode |

### Earnings mode permutation matrix

| Mode | Org Kind | Platform | Org Gets | Consultant Gets |
|------|----------|----------|----------|----------------|
| Marketplace | BUYER | 20% | Nothing | 80% |
| Agency split | PROVIDER/HYBRID | 10% (default) | 5% (default) | 85% (default) |
| Platform-only | PROVIDER/HYBRID | 100% | 0% | 0% |
| Internal consultant | PROVIDER/HYBRID | 10% | 90% | 0% |

---

### Prerequisites
- [ ] Set `ENABLE_PROVIDER_ORGS=true` in `.env`
- [ ] Run `npm run db:generate`
- [ ] Run `npx prisma db push` or `npx prisma migrate dev`
- [ ] Start dev server (`npm run dev`)

### Phase 1: Schema & type safety
- [ ] `npx prisma validate` passes
- [ ] `npx tsc --noEmit` passes (0 errors excluding pre-existing test file)
- [ ] `npx prisma db seed` runs without errors (creates PROVIDER + HYBRID test orgs)

### Phase 2: Org creation wizard
- [ ] BUYER org creation works (all 4 billing modes selectable)
- [ ] PROVIDER org creation sets `status = PENDING_VERIFICATION` (not ACTIVE)
- [ ] HYBRID org creation sets `status = PENDING_VERIFICATION`
- [ ] Kind selector (BUYER/PROVIDER/HYBRID) renders correctly
- [ ] PREPAID_UNLIMITED option appears in billing mode selector
- [ ] When `ENABLE_PROVIDER_ORGS=false`: PROVIDER/HYBRID creation returns 501

### Phase 3: Admin verification
- [ ] `POST /api/admin/organizations/[orgId]/verify` with `action: "APPROVE"` → status = ACTIVE
- [ ] `POST /api/admin/organizations/[orgId]/verify` with `action: "REJECT"` → status = DEACTIVATED
- [ ] Non-admin users get 403

### Phase 4: Consultant application/approval
- [ ] `POST /api/organizations/[orgId]/consultants/apply` creates PENDING membership
- [ ] Auto-approve (`autoApproveConsultants=true`) creates ACTIVE immediately
- [ ] `POST /api/organizations/[orgId]/consultants` with `action: "APPROVE"` → ACTIVE
- [ ] `POST /api/organizations/[orgId]/consultants` with `action: "REJECT"` → REMOVED
- [ ] Duplicate application → 409
- [ ] Non-consultant user → 403
- [ ] OrgAuditLog rows created for each action

### Phase 5: Checkout — all 4 billing modes
- [ ] TAG_ONLY: learner pays at checkout, payment tagged to org, standard 80/20 earnings
- [ ] SEAT_PACK: credit deducted from pool, `paymentMethod = ORG_CREDIT`, `isMockPayment = true`
- [ ] INVOICED_MONTHLY: no gateway call, `paymentMethod = ORG_INVOICED`, tagged for invoice
- [ ] PREPAID_UNLIMITED: no gateway call, `paymentMethod = ORG_PREPAID`, `amount = 0`, `isMockPayment = true`
- [ ] PREPAID_UNLIMITED + expired contract → checkout blocked with license expired error

### Phase 6: 3-way earnings split
- [ ] PROVIDER org consultant booking: ConsultantEarnings (85%) + OrganizationEarnings (5%) + platform (10%)
- [ ] Internal consultant (`earningsRecipient=ORGANIZATION`): ConsultantEarnings (0%) + OrganizationEarnings (90%)
- [ ] Custom per-consultant rate: `customConsultantPayoutRate` overrides org default
- [ ] Platform-only mode (rates 100/0/0): ConsultantEarnings (0%), NO OrganizationEarnings row (zero guard)
- [ ] Math check: `platformFee + orgShare + consultantShare = grossAmount` always

### Phase 7: Refund routing
- [ ] TAG_ONLY refund → gateway refund
- [ ] SEAT_PACK refund → credit pool restored
- [ ] INVOICED_MONTHLY refund → payment unbilled from invoice
- [ ] PREPAID_UNLIMITED refund → no financial action, `status = SUCCEEDED`
- [ ] 3-way split refund → both ConsultantEarnings + OrganizationEarnings reversed proportionally

### Phase 8: Payout pipeline
- [ ] `POST /api/organizations/[orgId]/payouts` creates batch from READY earnings
- [ ] Redis lock prevents concurrent batch creation (409)
- [ ] Minimum payout amount guard works
- [ ] `POST /api/admin/org-payouts/process` processes PENDING payouts
- [ ] Linked earnings marked as PAID after processing
- [ ] OrgAuditLog: PAYOUT_INITIATED + PAYOUT_PROCESSED

### Phase 9: Payout account encryption
- [ ] PUT payout account encrypts with AES-256-GCM (not base64)
- [ ] GET payout account never leaks full account number (only last4)
- [ ] OrgAuditLog: SETTINGS_CHANGED for PUT/DELETE

### Phase 10: HYBRID org
- [ ] Dashboard sidebar shows both BUYER + PROVIDER sections
- [ ] Learner booking uses BUYER billing mode (credit deduction)
- [ ] External client booking triggers 3-way earnings split
- [ ] Same-org learner books with same-org consultant: both financial streams activate independently

### Phase 11: Public pages
- [ ] `/org/[slug]` renders for ACTIVE PROVIDER/HYBRID org
- [ ] `/org/[slug]` returns 404 for BUYER org
- [ ] `/explore/companies` lists PROVIDER + HYBRID orgs only
- [ ] Consultant cards in `/explore/experts` show org badge with link

### Phase 12: Dashboard
- [ ] Payouts page: stat cards, payout history table, "Create Payout Batch" button
- [ ] Consultants page: active list, pending applications, approve/reject buttons
- [ ] Overview hub: PROVIDER stat cards (active consultants, total payouts)
- [ ] Billing page: shows "Prepaid Unlimited" mode label for PREPAID_UNLIMITED orgs

### Phase 13: Auth & error paths
- [ ] Outsider (non-member) → 403 on all org APIs
- [ ] ORG_LEARNER cannot create payouts (requires ORG_OWNER)
- [ ] Rate sum validation: rates not summing to 1.0 → 400

### Automated E2E test prompt
Full 11-phase, 40+ test case prompt at:
`prompts/enterprise-tests/e2e-enterprise-agent-002-billing-modes-provider-hybrid.md`

Uses Supabase MCP (CRUD) + Chrome DevTools MCP (UI) against `localhost:3000`.


teetangh:
ChatGPT 5.4 codex feedback

I reviewed this stacked PR by starting from the revamped `docs/enterprise/*` set and then walking the implementation through the provider/hybrid, billing, payout, public-page, and consultant lifecycle paths. The branch is much closer now, and the docs are dramatically clearer than before, but there are still a few high-signal gaps worth calling out before we call the stack fully merge-ready.

1. Multi-org provider attribution is still nondeterministic.

The docs say multi-org consultants use the “first active” PROVIDER/HYBRID membership, but `resolveOrgSplit()` currently does a plain `findFirst()` with no `orderBy`, and the explore/discovery badge path does the same thing with `take: 1` and no ordering. In practice that means a consultant who belongs to two active provider orgs can have earnings routed to one org on one request and show the other org’s badge on another request, depending on row order. That is both a financial correctness problem and a product-consistency problem.

Files to look at:
- `lib/payments/payouts/earnings-service.ts`
- `lib/data/explore-experts.ts`
- `app/api/user/consultants/route.ts`

Proposed fix: introduce one explicit source of truth for org precedence, for example `isPrimaryProviderOrg`, `displayPriority`, or at minimum a deterministic `orderBy` shared by both earnings routing and badge selection. The important thing is that earnings attribution and UI badge attribution should resolve the same org, every time.

2. Org payout processing is still not safely claim-once under concurrency.

`processOrgPayout()` reads the payout row, checks for `PENDING`, and then does a separate update to `PROCESSING`. Two admin/process jobs can both read the same row as `PENDING` before either update lands. On Razorpay you at least pass an idempotency key, but on Stripe the org branch currently calls `createTransfer()` without any comparable idempotency guard, so duplicate transfer attempts are still possible if this endpoint is triggered concurrently or retried badly.

Files to look at:
- `lib/payments/payouts/org-payout-service.ts`
- `app/api/admin/org-payouts/process/route.ts`
- `lib/payments/payouts/stripe-connect.ts`

Proposed fix: claim the payout atomically with a conditional update like `updateMany(where: { id, status: PENDING }, data: { status: PROCESSING })` and only continue when the affected row count is 1. After that, pass/persist a deterministic idempotency key for Stripe too, or otherwise make the transfer layer retry-safe.

3. Credit-pack purchase settlement now avoids double-crediting, but the state model is still overloaded in a way that will age badly.

The webhook now uses `OrgCreditPurchase.expiresAt` as a processed marker, while the orphan cleanup job uses the same field with an epoch sentinel to mean “cancelled/orphaned.” The docs still describe `expiresAt` as unused optional credit expiry and describe `paymentId` as the linkage to the funding payment, but the live implementation is now encoding three different meanings into one nullable field and still does not persist gateway settlement identifiers for the purchase.

Files to look at:
- `app/api/webhooks/utils.ts`
- `app/api/organizations/[orgId]/credits/purchase/route.ts`
- `scripts/enterprise/cleanup-orphaned-org-credit-purchases.ts`
- `prisma/schema.prisma`
- `docs/enterprise/09-credit-system.md`

Proposed fix: give `OrgCreditPurchase` explicit settlement fields such as `status`, `processedAt`, `cancelledAt`, `providerOrderId`, and `providerPaymentId`, then update webhook/cleanup/history logic to use those. That keeps the idempotency win without sacrificing future expiry semantics or audit clarity.

4. The consultant lifecycle is still more complete at the API/doc level than at the product level.

The docs say self-application happens from `/org/[slug]` and that admins review pending applications from the Consultants dashboard. But the public org page still has no actual apply action, and the dashboard consultants page only lists existing consultants; it does not expose pending applications, filters, or approve/reject controls. So the API exists, but the product flow is not actually navigable end to end yet.

Files to look at:
- `docs/enterprise/06-consultant-lifecycle.md`
- `app/org/[slug]/page.tsx`
- `app/dashboard/organization/[orgId]/consultants/page.tsx`
- `app/api/organizations/[orgId]/consultants/apply/route.ts`
- `app/api/organizations/[orgId]/consultants/route.ts`

Proposed fix: either wire the actual UI for apply/review, or narrow the docs/PR wording so we are not calling the lifecycle “implemented” before users can really do it from the product.

5. Per-consultant payout controls are still schema/business-logic only.

A lot of the new enterprise story depends on `customConsultantPayoutRate` and `earningsRecipient`, and the docs explicitly use them for “custom override” and “internal consultant” scenarios. But I could not find an API or dashboard surface that lets an org admin actually set those fields. Right now the earnings engine supports the scenarios, but the product does not expose them, which means a meaningful part of the provider/hybrid value prop still depends on manual DB edits or seeded data.

Files to look at:
- `prisma/schema.prisma`
- `lib/payments/payouts/earnings-service.ts`
- `app/dashboard/organization/[orgId]/consultants/page.tsx`
- `app/api/organizations/[orgId]/consultants/route.ts`

Proposed fix: add a member-level PATCH/admin action for consultant payout settings, or explicitly scope those scenarios as follow-up work in the docs and PR description.

If I were drawing the merge line, I would treat items 1 and 2 as true correctness bugs, and items 3 through 5 as the remaining “implemented vs actually complete product” gap. That is a much smaller list than earlier rounds, and it feels like the branch is getting close.

teetangh:
ChatGPT 5.4 codex feedback

This round is a lot tighter, and the progress is real.

Compared with the last review, the branch now looks much closer to merge-ready:
- deterministic multi-org resolution is in place for both earnings attribution and org badge selection
- org payout processing now does an atomic `PENDING -> PROCESSING` claim and Stripe got an idempotency key too
- the public org page now has a real consultant apply CTA
- the consultants dashboard now supports pending application review/approve/reject
- the member PATCH route now exposes the consultant payout-control fields instead of leaving them schema-only

At this point I’d call the stacked PR roughly “very close” rather than “still structurally incomplete.” I only see a short remaining list.

1. Local typecheck is failing again, so I would still treat this as the main merge-readiness blocker until it is explained or fixed.

`npx tsc --noEmit` fails locally on:
- `tests/typescript/race-conditions/test-checkout-race-condition-fix.ts:104`

The code is logging `result.value.paymentIntent.id` after only checking `result.status === "fulfilled"`, but TypeScript still sees `paymentIntent` as optional there. If CI is green, then the CI tsconfig/scope may differ from the local one; otherwise this is a real regression and should be cleaned up before merge.

2. The new per-consultant payout controls still need stricter domain validation.

`PATCH /api/organizations/[orgId]/members/[memberId]` now accepts `customConsultantPayoutRate` and `earningsRecipient`, which is good progress, but the handler still accepts those fields before checking that the target is actually an `ORG_CONSULTANT` in a `PROVIDER`/`HYBRID` org. It also does not validate the override against the org’s rates. Right now a bad override only gets caught later in `resolveOrgSplit()`, where negative `orgShare` is silently clamped to zero.

Files to look at:
- `app/api/organizations/[orgId]/members/[memberId]/route.ts`
- `lib/payments/payouts/earnings-service.ts`

Proposed fix: fail fast when these fields are present unless the target member is an org consultant in a provider-capable org, and validate the effective rate combination at write time instead of relying on runtime clamping.

3. Small product/UX follow-up: the new public apply CTA is functional, but it is still pretty raw for logged-out users.

`app/org/[slug]/ApplyButton.tsx` just POSTs immediately and surfaces whatever API error comes back. Since the public org page is now the documented self-application entry point, a gentler path would be to redirect unauthenticated users to sign-in, or at least show a clearer “sign in as a verified consultant to apply” message.

If I were drawing the line for merge: item 1 is the one I’d still treat as a real blocker, item 2 is the last important correctness hardening, and item 3 is a nice cleanup if you want the public flow to feel finished rather than merely wired.

teetangh:
ChatGPT 5.4 codex feedback

This round is mostly progress, and the branch feels very close.

Compared with the last pass:
- the local typecheck blocker looks fixed
- per-consultant payout controls now have proper domain/rate validation
- the public apply CTA has much better auth/onboarding UX
- the PROVIDER `billingMode = NULL` model is now reflected in the schema/API and the main sidebar nav

At this point I’m not seeing new deep payment/payout correctness bugs. The main remaining gaps are product-consistency follow-through from the new PROVIDER billing model.

1. The create-org wizard still treats billing as universal even though PROVIDER orgs now have `billingMode = NULL`.

`app/dashboard/organization/create/page.tsx` still always advances through a `BillingStep`, and on step 0 it still posts `billingMode: merged.billingMode ?? "TAG_ONLY"` even for PROVIDER orgs. `app/dashboard/organization/create/components/BillingStep.tsx` also still presents the four billing-mode choices as if they apply to every org kind. That means the product flow is now out of sync with the schema/docs: PROVIDER orgs are modeled as having no billing mode, but the wizard still asks the user to choose one and does not swap in the provider-specific revenue configuration step implied by the types/docs.

Files to look at:
- `app/dashboard/organization/create/page.tsx`
- `app/dashboard/organization/create/components/BillingStep.tsx`
- `app/dashboard/organization/create/types.ts`

Proposed fix: make step 1 conditional by org kind. BUYER/HYBRID can keep the billing step; PROVIDER should skip billing mode entirely and either show a provider revenue/payout setup step or just continue without billing language.

2. A couple of dashboard surfaces still assume every org has a visible billing mode, so PROVIDER orgs will show awkward empty/`null` UI.

`app/dashboard/organization/page.tsx` still types `billingMode` as always present and always renders a billing badge. With PROVIDER orgs now returning `null`, that badge becomes empty/useless. `app/dashboard/organization/[orgId]/settings/page.tsx` still types `billingMode` as `string` and renders a “Currently <strong>{billingMode}</strong>” card, which will read as `null` for PROVIDER orgs.

Files to look at:
- `app/dashboard/organization/page.tsx`
- `app/dashboard/organization/[orgId]/settings/page.tsx`

Proposed fix: make those views kind-aware. Either hide billing mode entirely for PROVIDER orgs or replace it with provider-relevant financial language (for example, payout cadence / revenue split instead of billing mode).

If I were drawing the line now: this no longer looks blocked on financial correctness. It looks blocked on a small amount of product cleanup so the UI fully matches the corrected enterprise model. That is a much better place to be than a few rounds ago.

teetangh:
## Race Condition & TOCTOU Audit — Enterprise Routes

Audited the entire enterprise API surface for production concurrency risks. Summary of what was found and fixed in commit 19da4448.

---

### Already Safe (no action needed)

| Concern | Why |
|---------|-----|
| Seat acquisition overdraft | Atomic `UPDATE … WHERE seatsUsed < seatsTotal` — loser gets 0 rows → 403 |
| Credit deduction overdraft | Same `UPDATE … WHERE balance >= amount` pattern in `deductCredits()` |
| Credit purchase webhook double-processing | `updateMany({ where: { status: "PENDING" } })` → `count=0` on retry, skips credit grant |
| Payout batch concurrency | `acquireLock()` wraps `createOrgPayoutBatch()` — already serialized |

---

### Fixed

#### 🔴 CRITICAL — Invitation double-accept (`invitations/accept/route.ts`)

**Before:** `invitation.status !== "pending"` check was pre-TX. Two concurrent accepts both passed the check, both entered the TX, both acquired a seat and created a member.

**After:** TX opens with `tx.invitation.updateMany({ where: { id, status: "pending" } })`. Only the first request gets `count=1`; the second gets `count=0` → `INVITATION_ALREADY_ACCEPTED` → full rollback (seat, member, audit log never written). Same pattern as the credit webhook idempotency guard.

---

#### 🔴 CRITICAL — Last-owner demotion/removal race (`[memberId]/route.ts` PATCH + DELETE)

**Before:** `countActiveOwners()` ran outside the TX. Org with 2 owners: concurrent demote-Alice + remove-Bob both saw count=2, both proceeded → 0 owners.

**After:** Guard moved **inside** both TX blocks as a *post-update* count using the `tx` client. Under READ COMMITTED: the TX that commits second sees its own uncommitted demotion + the first TX's committed demotion = 0 owners → throws `LAST_OWNER` → rollback. At least one owner always survives.

---

#### 🟠 HIGH — SSO domain claim TOCTOU (`sso/route.ts` + `schema.prisma`)

**Before:** `findMany(hasSome: domains)` conflict check then `upsert` — classic TOCTOU. Two orgs could both pass the check concurrently and both claim `acme.com`, making SSO routing ambiguous (wrong IdP for users).

**After:** Added `OrgDomainClaim` model with `domain @unique`. PATCH now runs `deleteMany` + `createMany` + settings `upsert` atomically in one TX. If another org holds a domain, `createMany` throws P2002 and the whole TX rolls back → 409. DB-level constraint, not application-level check.

> **Migration required:** `prisma migrate dev` creates the `org_domain_claims` table.

---

#### 🟠 HIGH — Duplicate pending invitations (`[orgId]/invitations/route.ts`)

**Before:** `findFirst(status: "pending")` then `create` — two concurrent POSTs for the same email both found no existing invite and both created one.

**After:** Wrapped in a `Serializable` transaction. PostgreSQL SSI detects overlapping read-write cycles and aborts the loser with P2034 → surfaced as 409 "already pending."

---

#### 🟠 HIGH — Billing mode change after first payment (`[orgId]/route.ts`)

**Before:** `payment.count()` check was outside the transaction (batch array `$transaction([...])`). A concurrent payment webhook could land between the count read and the profile update → org in wrong billing mode with a mismatched payment.

**After:** Converted to interactive `$transaction(async (tx) => {...})`. `tx.payment.count()` runs inside the same TX context so no payment can be inserted between the check and the update.

---

#### 🟡 MEDIUM — Admin org verification double-approve/reject (`admin/.../verify/route.ts`)

**Before:** `findFirst(status: "PENDING_VERIFICATION")` then `update` — two admins concurrently acting on the same org would race, with the second silently overwriting the first.

**After:** Single atomic `updateMany({ where: { status: "PENDING_VERIFICATION" } })`. `count=0` → 404. One action wins, the other gets a clear error.

---

#### 🟡 MEDIUM — `OrgCreditPurchase.providerOrderId @unique` (`schema.prisma`)

Razorpay could theoretically create two `OrgCreditPurchase` rows with the same `providerOrderId` via retried order-creation calls. The existing `status: PENDING` idempotency guard only protects within a single row. Added `@unique` to prevent two rows with the same order ID from both being credited.

> **Migration required:** Adds a unique index on `org_credit_purchases.providerOrderId`.

---

### No Redis locks needed

The booking algorithm needs Redis because slot *discovery* spans many queries requiring a consistent snapshot (hence the 3-layer: Redis lock → 120s Prisma TX → in-TX conflict validation). Enterprise operations are point-mutations on known rows — all critical sections fit in a single short `$transaction()`. Adding locks here would add latency and complexity without benefit.

teetangh:
## Finance Audit Fixes — Round 1

Self-review pass covering taxation, refunds, payouts, disputes, and org billing. All issues below are addressed in commit 726c52de.

---

### P0 — Critical (fixed)

**Invoice payment webhook race condition** (`app/api/webhooks/utils.ts`)

The previous `invoice_payment` handler did a non-atomic `findUnique` → `update`. Razorpay fires both `payment.captured` and `order.paid` for the same order; both events could pass the `status === "PAID"` check before either committed, risking a duplicate state transition. Replaced with:
```ts
prisma.organizationInvoice.updateMany({
  where: { id: invoiceId, status: { not: "PAID" } },
  data: { status: "PAID", paidAt: new Date(), providerPaymentId: razorpayPaymentId },
})
```
Same `updateMany`-with-guard pattern already used correctly for the `credit_purchase` branch.

**`providerPaymentId` never stored** (`OrganizationInvoice`)

The comment said "Store the Razorpay order ID for reconciliation" but the code stored nothing. Added `providerPaymentId String?` to `OrganizationInvoice` (schema + migration needed — see below) and populate it from the webhook.

---

### P1 — Important (fixed)

**`generate-invoice` was not applying GST** (`billing/generate-invoice/route.ts`)

Auto-generated INVOICED_MONTHLY invoices stored `taxAmount: null`, `taxRate: null`, `hsnCode: null`. The tax engine already existed — it just wasn't called here. Now calls `determineTax({ baseAmountPaise: subtotal, buyerCountry: "IN", serviceType })` using the dominant appointment type across unbilled payments. Invoice `amount` includes 18% GST; `taxAmount`, `taxRate`, `hsnCode` are populated.

**No OVERDUE invoice cron** (`app/api/cleanup/mark-overdue-invoices/route.ts` — new file)

Invoices past `dueDate` in `SENT` status were never promoted to `OVERDUE`. Added a daily cron (schedule: `01:00 UTC`) that does one bulk `updateMany WHERE status=SENT AND dueDate < now`.

**`creditRefund()` missing `memberProfileId`** (`lib/payments/operations/org-credits.ts`)

`deductCredits()` accepts `memberProfileId` for ledger attribution; `creditRefund()` didn't, making refund ledger rows anonymous. Signature extended to match.

**Domain claims not cleared on org deactivation** (`app/api/organizations/[orgId]/route.ts`)

`DELETE` only set `status = DEACTIVATED`; `OrgDomainClaim` rows remained, permanently blocking those email domains from being claimed by other orgs. Wrapped DELETE in a transaction that `deleteMany`s claim rows first.

---

### P2 — Compliance (fixed)

**GSTIN format not validated** (`billing/invoices/route.ts`)

`gstin: z.string().optional()` accepted anything. Added the standard 15-char regex (`/^\d{2}[A-Z]{5}\d{4}[A-Z][1-9A-Z]Z[0-9A-Z]$/`).

**Audit trail gaps for financial events**

`OrgAuditAction` had `PAYOUT_INITIATED`/`PAYOUT_PROCESSED` but no financial event types. Added `INVOICE_GENERATED`, `CREDITS_PURCHASED`, `INVOICE_PAID` to the enum. Emit them from: `generate-invoice` (inside TX), `credits/purchase` (fire-and-forget), and the invoice payment webhook handler.

---

### Schema migration required

```sql
ALTER TABLE "OrganizationInvoice"
  ADD COLUMN IF NOT EXISTS "providerPaymentId" TEXT;

ALTER TYPE "OrgAuditAction" ADD VALUE IF NOT EXISTS 'INVOICE_GENERATED';
ALTER TYPE "OrgAuditAction" ADD VALUE IF NOT EXISTS 'CREDITS_PURCHASED';
ALTER TYPE "OrgAuditAction" ADD VALUE IF NOT EXISTS 'INVOICE_PAID';
```

Run `npx prisma migrate dev --name enterprise_finance_fixes` to generate and apply.

---

### Still deferred (tracked)

| Item | Issue |
|------|-------|
| Monthly invoice auto-generation cron (Inngest) | #661 / Phase K |
| Invoice PDF rendering (`pdfUrl`) | #438 |
| Credit notes for post-payment refunds on paid invoices | #655 |
| Stripe support for org credit purchase + invoice pay | Phase L |

teetangh:
ChatGPT 5.4 codex feedback

This branch is now very close.

Progress since the earlier rounds is strong:
- `npx tsc --noEmit` passes locally again
- the wizard now branches correctly for BUYER / PROVIDER / HYBRID
- the `billingMode = NULL` PROVIDER model is reflected across schema, API, settings, org list, and sidebar nav
- credit-purchase settlement now has explicit status fields instead of overloading `expiresAt`
- SSO domain claims, invitation acceptance, owner demotion, and invoice payment webhook races all look materially better and more intentional

I only see one remaining issue worth calling out before I’d call this fully merge-ready:

1. Invoice settlement is now idempotent, but the allowed state transition is still too broad.

`app/api/organizations/[orgId]/billing/invoices/[invoiceId]/pay/route.ts` still allows paying any invoice that is not `PAID` or `CANCELLED`, which includes `DRAFT`. Then `app/api/webhooks/utils.ts` marks the invoice as `PAID` with `where: { id: invoiceId, status: { not: "PAID" } }`, which means the webhook will also settle invoices from states other than `SENT` / `OVERDUE`.

The dashboard UI only shows a Pay button for `SENT` / `OVERDUE`, so the product contract already implies those are the payable states. The API/webhook should enforce the same state machine; otherwise a direct API call (or a stale gateway order created before some later invoice-state change) can bypass the intended invoice lifecycle.

Proposed fix:
- In `billing/invoices/[invoiceId]/pay/route.ts`, require `invoice.status` to be one of `SENT` or `OVERDUE`.
- In `handleOrgPaymentSuccess()` for `invoice_payment`, change the guard to only update invoices whose status is in `SENT` / `OVERDUE`.

At this point I’d put the branch at the high-90s for merge readiness. This is no longer a “many things missing” situation — it’s one last state-machine hardening pass.

teetangh:
Fixed in a18a6390.

**`billing/invoices/[invoiceId]/pay/route.ts`** — replaced the two separate PAID/CANCELLED checks with a single positive guard: rejects anything whose status is not `SENT` or `OVERDUE`, with a distinct error message for each disallowed state (DRAFT, PAID, CANCELLED, and a fallback for any future states).

**`webhooks/utils.ts` `handleOrgPaymentSuccess`** — changed the `updateMany` WHERE from `status: { not: "PAID" }` to `status: { in: ["SENT", "OVERDUE"] }`. Both the pay route and the webhook now enforce the same state machine contract: only SENT/OVERDUE invoices can transition to PAID.

Issue #780, opened 2026-05-31
Enterprise — money type integrity: migrate all money columns Int→BigInt (number at JS boundary)


teetangh:
## Why this issue

Extracted from **#778 §A** (v3 mega-audit) as its own work item — the change is large enough (touches ~71 columns + ~180 aggregation sites) that folding it silently into the v0 lockdown PR would be unreviewable. This issue carries a **complete, proven implementation plan** (the schema conversion script + the Prisma boundary extension both ran green locally) plus the integration findings from a spike, so the next session can execute cold.

**The defect (confirmed in code):** nearly every money column is Postgres `Int` (int4), ceiling **2,147,483,647 paise = ₹2.14 crore**. Only `LedgerEntry.amountPaise` is `BigInt`. A single enterprise annual contract or an aggregated monthly invoice above ₹2.14cr **silently overflows today**. This is the #1 finding of the v3 audit.

**Pre-MVP framing:** DB is wiped before launch, so there is no backfill — this is a pure schema-type change plus the code plumbing it forces. Because v0 is the schema-freeze PR, **this should land before v0 merges** (changing a money column type after the freeze = a migration on real data). See "Sequencing".

## Locked decision (from #778)

**Money is `BigInt` at the DB column, `number` at the JS boundary.** Rationale: the overflow is a *Postgres int4* problem, not a JS one — `Number.MAX_SAFE_INTEGER` = 9,007,199,254,740,991 paise ≈ **₹90 lakh crore**, far beyond any invoice. So we widen the column to int8 (kills overflow) and convert `bigint→number` once at the Prisma read boundary via a client extension. The entire app / API / Razorpay payloads / Zod / frontend stay `number` and untouched. Writes are unaffected — Prisma accepts `number` for `BigInt` inputs.

Rejected alternatives: `bigint` end-to-end (forces a sweep of every arithmetic/serialization/Razorpay/Zod/frontend site, no domain benefit under ₹90 lakh crore); `string` at boundary (breaks all numeric math + Razorpay).

---

## Step 1 — Schema: flip 71 money columns `Int`→`BigInt`

[...]

Issue #779, opened 2026-05-31
Enterprise v4 mega-audit — lifecycle, mutation appropriateness & customer-state experience (PM lens)


teetangh:
# Enterprise v4 mega-audit — lifecycle, mutation appropriateness & customer-state experience (PM lens)

Fifth in the series — **#768 (v0 schema)**, **#776 (v1 backend/money)**, **#777 (v2 frontend)**, **#778 (v3 schema accuracy & finance)**, and now **v4, from a principal-PM lens**: can *any* customer, in *any* state-permutation, reach their desired outcome — without getting stuck, silently orphaned, or shown a generic screen when they need to act? This audit is about **appropriateness and lifecycle**, not "is the button wired." Grounded in three PM Explore passes + benchmarking (Stripe/Ramp/Vercel/WorkOS/Atlassian/GitHub/Rippling).

## The thesis
The infrastructure is mature; the gap is **product discipline**. Today the system shows the user *what our APIs return*, not *what they need to do*. Money gets *silently* stuck (recoverable, but the customer is never told). Dangerous mutations are *allowed* (flip off hosting and orphan your experts). And entities become *zombies* (assignments that expired but still count). v4 fixes the experience so a customer is **never silently stuck and always sees their next action.**

## ⚠️ Sequencing (like v3, v4 partially folds back into v0/v1/v2)
- **Schema bits → v0 (before freeze):** `ProgramAssignment.status` enum, `Contract.autoRenew`, `Program.configLockedAt`, `Refund.failureReason`, invoice dunning timestamps, an SSO break-glass field, contract-supersession link. `[v0-BLOCKER]`
- **Guards + crons + logic → v1:** capability/funding/config/offboarding guards, the cycle engine, dunning, stuck-state timeouts, break-glass, notifications.
- **Surfaces → v2:** the action-center dashboard, per-state banners, self-serve verification, dispute/overage/collection views, confirmation dialogs.
- **Dependency flag:** the **cycle engine (§B) must land before** the "applies next cycle" money-config edit (§A) — don't ship the edit button without the cycle it depends on.

## Tags
`[P0]`/`[P1]`/`[P2]` priority · `[SCHEMA→v0]` schema, fold into freeze · `[GUARD]` mutation guard · `[CRON]` background job · `[UI]` surface · `[NOTIF]` notification · `(verify)`.

## Locked design decisions
1. Capabilities: additive free, **removal guarded** (wind-down required).
[...]

Issue #776, opened 2026-05-29
Enterprise v1 mega-audit — path to full enterprise completeness (SPONSOR/HOST/HYBRID + cross-cutting + pre-MVP architecture)


teetangh:
# Enterprise v1 mega-audit — path to full enterprise completeness

Sibling to the v0 mega-audit (#768). v0 froze the schema and built the money *engine*; **v1 is about making the enterprise subsystem actually serve every customer archetype our model claims to support, end-to-end, and making the cross-cutting seams real.** This is the pre-MVP moment to also decide the big architectural questions while changing them is still cheap (no prod data, no backfill).

### Scope & framing
- **In scope:** everything needed to onboard and fully serve SPONSOR / HOST / HYBRID / PERSONAL-funded orgs on the *current* model.
- **Explicitly OUT of scope (Programs v2 — separate epic):** PROJECT / RETAINER program types, reseller/affiliate orgs, multi-currency wallets (#711). These are deliberately excluded.
- **Already shipped in v0 lockdown — NOT relitigated here:** double-entry ledger cutover (#772), overage settlement incl. CHARGE_ORG / CHARGE_MEMBER / BLOCK + CREDIT_POOL paise money-meter (#775/#715/#753), invoice leg→rollup (#749), reachable-funding-path gate, webhook signing-key rotation, TDS 194-O rate fix (#737/#738). If a stale doc/issue lists these as pending, it's stale.

### Priority legend
`[P0]` correctness / launch-blocker for a segment we claim to serve · `[P1]` required for a real paying customer in that segment · `[P2]` first-90-days · `[P3]` polish / scale · `[TRIM]` cut or quarantine · `[ARCH]` architectural decision, sized separately.

---

## 0. Customer archetypes & their gating gaps

The honest "can we sell to them today?" matrix. Each archetype and the specific blockers standing between it and go-live:

- **SPONSOR · WALLET (prepaid credit pool)** — *Closest to ready.* Booking, wallet debit, overage, reconcile all wired. Blockers: live-payout disbursement is irrelevant (no payouts), but **onboarding still needs manual contract/program setup** (§F) and `ENABLE_HOST_ORGS`/funding UX polish. **Sellable after §F + §K.**
[...]

PR #785, opened 2026-06-01
feat(enterprise): #776 v1 mega-audit — money core, reversal engine, credit notes, disputes, observability

### LINKED ISSUE #716: Refunds + payouts + pricing subsystem — full implementation plan (consolidation epic)
## TL;DR

Refunds, partial refunds, payout reversals, credit notes, and the pricing-snapshot subsystem have grown organically. Several paths handle "partial" inconsistently — `refundEarnings` prorates earnings via `refundRatio`, but the cap reversal was always-full pre-PR-1e (and even post-PR-1e is patched in only at the cap layer, not unified across the whole subsystem). Payout reversal once `OrganizationPayout.status='COMPLETED'` is unimplemented; credit notes are schema-only; SUBSCRIPTION partial-delivery refunds don't exist.

This issue is the consolidation epic that unifies refund handling and ships the missing pieces (payout clawback, credit notes, refund state machine, pricing-snapshot consistency).

Cross-refs: pairs with #715 (overage charging). Started by PR-1e proportional-cap-reversal patch (commit `<TBD>`).

---

### Implemented and consistent

- **Consultant earnings refund** — `refundEarnings(refundAmount, paymentAmount)` computes `refundRatio = refundAmount / paymentAmount` and prorates `consultantShare` per earning row (`lib/payments/payouts/earnings-service.ts:583-589`). Caps the prorated amount at the remaining-reversible balance to prevent over-refund.
- **Org earnings refund** — same path, prorates `orgSharePaise` (`lib/payments/payouts/earnings-service.ts:619-625`).
- **Wallet credit on refund** — `walletCredit(reason='REFUND')` writes WalletEntry + FundingLedgerEntry inside the refund webhook tx (`app/api/webhooks/utils.ts:597-650`). Idempotent via the `WalletEntry.providerPaymentId` check.
- **TDS reversal (PAID earnings only)** — when a refund hits a PAID earning under `forceRefund=true`, a negative TDSRecord row is written (`lib/payments/payouts/earnings-service.ts:680-697`). Reversal proportional to `refundRatio`.
- **Status transition guard** — `assertEarningStatusTransitionLegal` (PR-1b / LED-2) blocks PAID → anything-other-than-REFUNDED at app layer (`lib/payments/payouts/earning-status.ts`).
- **Cap reversal proportional** — *PR-1e adds this* (Phase A); pre-PR-1e it was always full, post-PR-1e it accepts an `engagementsToReverse` arg.

### Implemented but inconsistent

- **Pricing snapshot location split.** Org-funded bookings persist `priceAtBookingPaise` on `BookingUtilization` (rate-card snapshot); personal bookings persist `originalAmount` on `Payment`. Two stores, two semantics — reconcile ledger has to know which to read per booking.
- **Refund-cap reversal coupling.** The cap reversal (`reverseBookingUtilization`) is only called from a subset of refund paths: the Razorpay refund webhook (`app/api/webhooks/utils.ts`) calls it; admin-initiated refunds (where they exist) may not. No central `RefundContext` orchestrates "always reverse cap when refunding an org-funded booking".
- **`refundEarnings` vs cap reversal sequencing.** Today these are separate calls in the same webhook tx. PR-1e Phase A wires the refundRatio across both, but the API is two helpers with two arg shapes — should converge to one.

### Stubbed / schema-only

- `OrganizationCreditNote` — model exists in `prisma/schema.prisma`; **no issuance API, no PDF route, no invoice-link wiring.**
- **Payout reversal / clawback** — once an `OrganizationPayout.status='COMPLETED'`, refunding the underlying earnings marks the earning row `REFUNDED` but the payout's `netPayoutPaise` is now stale. The reconcile invariant G (PR-1c) flags this drift but cannot remediate. **No clawback API; no negative payout writer; no consultant-side "we owe you less next cycle" path.**
- **Per-engagement SUBSCRIPTION pricing.** SUBSCRIPTION pays `priceAtBookingPaise = full price` on first allocation, 0 on subsequent (verified at `utils/slotAllocation/SlotAllocationService.ts:1530`). If the consultant only delivers 6 of 12 contracted sessions, the learner has overpaid; **no proration on partial delivery.**

### Operational gap

- Refund webhooks today are mostly fire-and-forget on the audit / Novu side. The ledger writes are transactional (PR-1b); the auxiliary writes are not. **A full refund-state-machine with retry + terminal states is not in place.**

---

### Phase 1 — Partial-refund unification

Every refund path (consultant earnings, org earnings, cap, wallet credit, TDS) accepts the same `(refundAmount, paymentAmount)` pair and applies a consistent `refundRatio`. PR-1e starts this for cap reversal; full unification is its own PR. Introduce a single `RefundContext` type:

```typescript
export type RefundContext = {
  paymentId: string;
  refundAmountPaise: number;
  paymentAmountPaise: number;
  refundRatio: number; // computed: refundAmount / paymentAmount
  isFull: boolean;     // refundRatio === 1
  reason: "GATEWAY" | "ADMIN" | "CHARGEBACK" | "OVERAGE_REVERSAL" | "PAYOUT_CLAWBACK";
  initiatedBy: { kind: "USER" | "ADMIN" | "SYSTEM"; id?: string };
};

export async function processRefund(tx, ctx: RefundContext): Promise<void> {
  // Sequenced: ledger writes first, side-effects after
  await reverseEarnings(tx, ctx);
  await reverseCapUtilization(tx, ctx);
  await creditWallet(tx, ctx);          // for org-funded paths only
  await reverseTdsIfPaidOut(tx, ctx);   // forceRefund path
  await maybeIssueCreditNote(tx, ctx);  // for INVOICED_MONTHLY mid-cycle
  await emitRefundAuditEvent(tx, ctx);
}
```

Every existing call site (Razorpay refund webhook, admin refund endpoints) collapses to one call. Cross-cutting tests pin the sequencing and the rounding semantics.

### Phase 2 — Payout reversal / clawback

When a refund hits an earning whose payout is `COMPLETED`, the payout's `netPayoutPaise` is now wrong. Three options, each with tax / accounting implications:

- **(a) Negative payout next cycle.** Issue an `OrganizationPayout` row with negative `netPayoutPaise` that adjusts the next batched payout. Cleanest from an accounting standpoint (the books always net out); requires a "minimum payout = max(0, computed)" floor for the next batch.
- **(b) Hold the consultant from receiving the next payout until the clawback is settled.** `ConsultantEarnings` refund creates a `ClawbackHold` row that the next payout cron consults; if positive, it reduces that cycle's earnings before batching.
- **(c) Forced repayment via wallet debit.** Heavy-handed; requires the consultant to have a wallet with funds. Probably never the right answer.

Recommendation: ship (b) first (smaller surface), evolve to (a) once the rhythm is established. Add `ConsultantEarnings.clawbackPending Boolean` flag + `Clawback` table.

### Phase 3 — Credit notes

Issue `OrganizationCreditNote` rows when an INVOICED_MONTHLY org refunds within the same billing cycle. PDF generation mirrors the invoice PDF (`lib/pdf/invoice-renderer.tsx`). API:

- `POST /api/organizations/[orgId]/credit-notes` — admin-driven (rare but needed).
- Auto-issue from `processRefund` when `OrganizationInvoice.status='ISSUED' AND ctx.refundAmountPaise > 0`.
- PDF route: `GET /api/organizations/[orgId]/credit-notes/[creditNoteId]/pdf` (cache via Supabase signed URL like invoice PDFs).

### Phase 4 — Pricing snapshot consistency

Every paid path writes `priceAtBookingPaise` to a consistent location. Today:
- Org-funded: `BookingUtilization.priceAtBookingPaise` (right place — captures rate-card snapshot).
- Personal-funded: `Payment.originalAmount` (different field).
- The reconcile ledger has to read the right one based on whether the booking is org-funded.

Options:
- **(a) Unify on `BookingUtilization`** for every booking (create a row even for personal bookings). Mirrors the wallet/funding-ledger split (one canonical store). Migration risk: existing personal bookings lack `BookingUtilization` rows.
- **(b) Document the split** and add a unified read helper `getBookingPriceSnapshot(bookingId)`. Smaller surface.

Recommendation: (b) for v1; revisit (a) when org bookings outnumber personal bookings.

### Phase 5 — Per-engagement pricing for SUBSCRIPTION

Today SUBSCRIPTION pays full price on first allocation and 0 on subsequent. If the consultant only delivers 6 of 12 sessions, the learner has overpaid. Spec the prorated-on-non-delivery refund flow:

- New cron `jobs/refunds/sweep-undelivered-subscriptions.ts` runs at end-of-cycle.
- For each `Payment` with `appointmentType='SUBSCRIPTION'` whose `Subscription.endDate < now()`:
  - Count `actualDeliveredSessions = COUNT(Appointment WHERE status='COMPLETED' AND subscriptionId=X)`.
  - Compute `expectedRefundRatio = (totalSessions - actualDeliveredSessions) / totalSessions`.
  - If `expectedRefundRatio > 0`, auto-trigger `processRefund` with `reason='UNDELIVERED'`.
- Manual override: admin endpoint to mark a subscription as "honored despite missing sessions" (e.g., consultant offered free extras).

### Phase 6 — Refund SLA + queue

Refund webhooks today fire-and-forget on the funding/audit/Novu side. A retry queue with explicit terminal states:

- New `RefundJob` table with `status: PENDING | PROCESSING | SUCCEEDED | FAILED_RETRYABLE | FAILED_TERMINAL`.
- Existing webhook handler enqueues a RefundJob and commits; a worker picks it up and runs `processRefund`.
- Failed jobs retry with exponential backoff (max 5 attempts). Terminal failures alert finance via Novu.

---

## Schema additions to consider

```prisma
model RefundJob {
  id                String          @id @default(uuid())
  paymentId         String
  refundAmountPaise Int
  paymentAmountPaise Int
  reason            RefundReason
  status            RefundJobStatus @default(PENDING)
  attempts          Int             @default(0)
  lastError         String?
  createdAt         DateTime        @default(now())
  completedAt       DateTime?
}

enum RefundReason { GATEWAY ADMIN CHARGEBACK OVERAGE_REVERSAL PAYOUT_CLAWBACK UNDELIVERED }
enum RefundJobStatus { PENDING PROCESSING SUCCEEDED FAILED_RETRYABLE FAILED_TERMINAL }

model Clawback {
  id                  String   @id @default(uuid())
  consultantProfileId String
  amountPaise         Int
  reason              String
  resolvedInPayoutId  String?
  createdAt           DateTime @default(now())
  resolvedAt          DateTime?
  @@index([consultantProfileId, resolvedAt])
}

model ConsultantEarnings {
  // ... existing
  clawbackPending     Boolean @default(false)
}

model OrganizationCreditNote {
  // ... existing
  invoiceId           String?  // link back to original invoice
  invoice             OrganizationInvoice? @relation(fields: [invoiceId], references: [id])
}
```

---

## Edge cases

- **Refund larger than original payment.** Impossible at gateway level but worth defending. `processRefund` clamps `refundAmount = min(refundAmount, paymentAmount - alreadyRefunded)`.
- **Multiple partial refunds totalling more than original.** `Payment.refundedAmountPaise` capped at `paymentAmount`; further refunds rejected with `409 OVER_REFUND`.
- **Refund initiated by org (admin-driven) vs by gateway (chargeback).** Different audit-log actions, same financial path. `RefundContext.reason` discriminates.
- **Refund of a payment whose org is now DEACTIVATED.** Refund still proceeds; the org's status gates new bookings, not historical reversals.
- **Concurrent refunds on the same payment.** Row-lock via `SELECT FOR UPDATE` on Payment in `processRefund`. Reconcile invariant catches drift.
- **Currency mismatch** (rare today, INR-only). When multi-currency lands, refund must use the same currency as the original payment.

---

## Acceptance criteria

- All refund paths share one `RefundContext` type and call `processRefund`.
- Reconcile invariants extended:
  - **J**: For every COMPLETED OrganizationPayout, `sum(attached earnings' consultantShare - refundedAmount) == netPayoutPaise` post-clawback adjustment.
  - **K**: For every PAID invoice with refunds, a matching CreditNote exists (or auto-issued).
  - **L**: For every Payment with `refundedAmountPaise > 0`, all its `BookingUtilization` rows have `reversedAt != null`.
- Credit notes generated automatically for INVOICED_MONTHLY mid-cycle refunds.
- Undelivered-subscription cron green; demo refund triggers correctly.
- Tour stop documenting the refund flow end-to-end.

---

## Estimate

~3000-4000 LOC across 6 phases. Probably ships as 3 PRs:
- **PR-A**: Phase 1 (RefundContext unification) + Phase 4 (pricing snapshot helper) — foundational.
- **PR-B**: Phase 2 (clawback) + Phase 6 (RefundJob queue) — financial state machine.
- **PR-C**: Phase 3 (credit notes) + Phase 5 (undelivered subscriptions) — features.

Filed alongside PR-1e in the engagement-counter audit. Pairs with #715 (overage charging).

### LINKED ISSUE #726: feat(enterprise): plan visibility enum (PUBLIC/ORG_ONLY/ORG_AND_PUBLIC) + marketplace leak guard
## Context

A consultant or org operator should be able to publish a plan (consultation / subscription / webinar / class) with one of three audiences:

- `PUBLIC` — visible in the global marketplace; anyone can book
- `ORG_ONLY` — visible only to ACTIVE members of the owning org; private internal training
- `ORG_AND_PUBLIC` — visible in BOTH the global marketplace AND the owning org's curated catalog (e.g., a thought-leadership webinar that doubles as employee training)

Today the schema has only the binary `Plan.organizationId String?` (null = personal/global, set = owned by org X). There is no `visibility` enum.

**Marketplace exposure:** the four marketplace query routes (`/api/plans/{consultations,subscriptions,webinars,classes}/route.ts`) do NOT filter on `organizationId` — `where = consultantId ? { consultantProfileId: consultantId } : {}`. So any org-owned plan that gets created later (via the `/api/organizations/[orgId]/plans/**` routes once they un-stub from #684) would leak into public results.

This issue locks down the schema + adds the immediate marketplace leak guard. UI work (consultant planner selector, org-side plans page, learner-side discovery) is out of scope here — those belong to #684.

## Schema migration

Add a new `Visibility` enum + `visibility` field to all four plan models in `prisma/schema.prisma`:

```prisma
enum Visibility {
  PUBLIC          // global marketplace
  ORG_ONLY        // private to ACTIVE members of owning org
  ORG_AND_PUBLIC  // marketplace + org catalog
}

model ConsultationPlan {
  // existing fields…
  visibility Visibility @default(PUBLIC)

  @@index([visibility])
  @@index([organizationId, visibility])
}
// same on SubscriptionPlan, WebinarPlan, ClassPlan
```

Backfill: every existing row is `organizationId IS NULL` today (consultant planner forms hardcode null at `EventPlannerForClass.tsx:299` and `EventPlannerForWebinar.tsx:294`), so default `PUBLIC` is correct without a data backfill.

App-layer invariant (no Postgres CHECK for now): `visibility != 'PUBLIC' ⇒ organizationId IS NOT NULL`. The Zod validator on the org-plans POST route should enforce it when #684 lands.

## Marketplace leak guard

Update `where` in the four routes:

- `app/api/plans/consultations/route.ts:18`
- `app/api/plans/subscriptions/route.ts:19`
- `app/api/plans/webinars/route.ts:107, 139, 150`
- `app/api/plans/classes/route.ts:113, 150, 161`

For unauthenticated / non-org-member callers:

```ts
where: { ...existing, visibility: { in: ['PUBLIC', 'ORG_AND_PUBLIC'] } }
```

For authenticated callers, add their org-only plans via OR. Read `session.user.organizationMemberships` (already populated by `customSession` in `lib/auth.ts:497`) and build:

```ts
where: {
  ...existing,
  OR: [
    { visibility: 'PUBLIC' },
    { visibility: 'ORG_AND_PUBLIC' },
    { visibility: 'ORG_ONLY', organizationId: { in: myActiveOrgIds } },
  ],
}
```

## Out of scope (defer)

- Consultant planner UI selector for visibility (the forms hardcode `organizationId: null` today)
- Org-side plans page `/dashboard/organization/[orgId]/plans` — #684
- Org-side learner catalog `/dashboard/organization/[orgId]/catalog` — Phase 2 of #684
- Checkout enforcement (rejecting a non-org-member trying to book an `ORG_ONLY` plan) — needs the UI to actually create such plans first
- Wiring `Organization.enforceOrganizationPlans` (field exists; no read site today)

## Acceptance

- [ ] `npx prisma validate` clean against the new schema
- [ ] `npx prisma db push` applies the enum + 4 columns + indexes against staging
- [ ] Marketplace `/api/plans/*` endpoints return zero plans with `visibility = ORG_ONLY` for anonymous callers
- [ ] An authenticated org member sees their org's `ORG_ONLY` plans (sanity test via seed fixture)
- [ ] Existing personal plans read back with `visibility = PUBLIC` (no data backfill needed)
- [ ] Unit test on the where-clause builder covering anonymous / member / non-member access scopes
- [ ] `npx tsc --noEmit` + `npm run test` clean

## References

- #684 — org plans page rehydration (the UI surface this unlocks)
- #674 — personal vs org scope split (broader pattern)
- #686 — v1 scope audit
- #727 — companion issue (referrals/trials/collaborators schema future-proofing)

teetangh:
# Enterprise v1 mega-audit (#776) — backend/money completeness

Single stacked branch off `feature/enterprise`. I validated the audit's claims against code first (it warns *"several bugs turned out to be already-fixed"*) — most of Wave-1 P0 was already shipped, so this PR delivers the **genuinely-pending** slices: correctness/compliance close-out, both `[ARCH]` decisions, credit-notes + disputes, and the live-payout go-live runbook.

**Part of #776 · Closes #726 · Closes #716**

> Schema deltas land as edits (no migration generated); `prisma migrate dev` is run once against the fresh DB. No backfills (pre-MVP reset).

## Verification
- `prisma validate` ✅ · `tsc --noEmit` ✅ 0 errors · `jest __tests__/enterprise/` ✅ **46 suites / 337 tests**, no regressions.
- DB-bound checks to run on reseed: `scripts/reconcile/reconcile-ledgers.ts` (expect `ok:true`, 0 findings) and `scripts/smoke/org-payout-sandbox-smoke.ts` (passes with `ENABLE_LIVE_PAYOUTS` unset).

> The original v1 mega-audit slices (sections A–E below) plus this session's **money-audit remediation round** (refund-cascade unification, concurrency hardening, GST/TDS fixes, PR-review resolutions) are documented further down under **"v1 money-audit — remediation round"**.

---

## A · Correctness & compliance close-out

**#726 — marketplace ORG_ONLY leak**
- `lib/data/explore-programs.ts` — curated carousels (`getCuratedPrograms`, 6 query blocks) + category counts (`getTopicsWithCount`) bypassed the visibility filter the API routes already apply. Now spread `marketplaceVisibilityWhere()` so `ORG_ONLY` plans never surface on `/explore`.

**MSME 43B(h) on the consultant/SELF payout path**
- `prisma/schema.prisma` — add `ConsultantPayout.mustPayByDate`.
- `lib/payments/payouts/payout-service.ts` — stamp via `computeMsmePaymentDeadline` using the consultant's *own* MSME status (the supplier on this path). Previously only the org path had it.
- `jobs/compliance/msme-payment-alerts.ts` — the daily sweep now covers **both** `OrganizationPayout` and `ConsultantPayout`, merged into one deadline-sorted alert.

**Observability sink (§K)**
- `lib/observability/betterstack-telemetry.ts` (new) — `emitTelemetryLog()` POSTs to BetterStack Telemetry (Bearer source token), best-effort, never throws.
- `lib/enterprise/system-events.ts` — `recordSystemEvent`/`recordSystemError` fire-and-forget to the sink after the DB write (DB row stays source of truth).
- `lib/feature-flags.ts` — `ENABLE_BETTERSTACK_TELEMETRY` (off by default).
- Callsites: `jobs/reconcile/reconcile-ledgers.ts`, `jobs/payouts/handle-stuck-payouts.ts`, `jobs/cleanup/dispatch-outbound-webhooks.ts` (queue backlog), `app/api/webhooks/razorpay/route.ts` (HMAC failure).

## B · `[ARCH]` Ledger balance snapshots
- `prisma/schema.prisma` — `LedgerAccountBalance` (1:1 with `LedgerAccount`, signed `balancePaise` + `entrySeq` guard).
- `lib/payments/ledger/post.ts` — `postLedgerTxn` folds each posting's signed delta into the snapshot inside the same tx (idempotency fast-path returns before any mutation → no double-apply). `ledgerBalancePaise` reads the O(1) snapshot; `ledgerBalanceFromJournalPaise` is the authoritative fallback. Replaces the O(n) `groupBy` scan.
- `scripts/reconcile/reconcile-ledgers.ts` — new `LEDGER_BALANCE_SNAPSHOT_DRIFT` invariant.

## C · `[ARCH]` Reversal engine + CLASS refund
- `lib/payments/operations/reversal-engine.ts` (new) — `applyReversal()` is the single front door for all reversals (`BOOKING` / `OVERAGE` / `PAYOUT_CLAWBACK` / `CLASS_MULTI`) sharing one idempotent ledger counter-posting. `BOOKING`/`OVERAGE` delegate to the proven `applyRefundCascade`; **`CLASS_MULTI` is new** — fans a single logical refund proportionally across the child payments of a consolidated CLASS purchase (no single `paymentId`), last child absorbing the remainder.
- `scripts/reconcile/reconcile-ledgers.ts` — new `REFUND_BOOKING_COHERENCE` invariant (fully-refunded payment ⇔ reversed `BookingUtilization`).

## D · Credit notes + org disputes/chargebacks

**Credit notes (GST Sec 34 / Rule 53)**
- `prisma/schema.prisma` — `OrgCreditNoteCounter` (independent FY-gapless CN series).
- `lib/payments/billing/credit-note-numbering.ts` (new) — `<PREFIX>-CN-<FY>-<seq>` via Prisma `upsert`+`increment` (no raw SQL).
- `lib/payments/operations/refund.ts` — mints a `CreditNote` on invoiced (`INVOICE_ACCRUAL`/CHARGE_ORG) refunds, referencing the original invoice with a proportional CGST/SGST/IGST split (replaces the `#716` TODO).

**Org dispute surface + chargeback money-path**
- `app/api/organizations/[orgId]/disputes/route.ts` (new) — finance-read list of disputes on org-funded payments.
- `app/dashboard/organization/[orgId]/disputes/page.tsx` (new) + `app/dashboard/organization/[orgId]/layout.tsx` — "Disputes" nav under Commerce (finance-gated).
- `app/api/webhooks/utils.ts` — `handleDisputeUpdated` LOST/CHARGE_REFUNDED now runs `applyOrgChargeback`: **org-wallet-first** (funder bears the chargeback), falling back to `ORG_RECEIVABLE`; balanced ledger counter-post + audit row.

## E · Live-payout go-live (gated)
`ENABLE_LIVE_PAYOUTS` stays **OFF**.
- `docs/enterprise/45-live-payout-go-live-runbook.md` (new) — pre-flip checklist, sandbox proof, canary, rollback.
- `scripts/smoke/org-payout-sandbox-smoke.ts` (new) — asserts the gate holds (no gateway submission while flagged off).

## Docs
`ENTERPRISE_SUBSYSTEM_CHECKLIST.md` (rows flipped), `ENTERPRISE_VERIFICATION_GUIDE.md` (v1 verifiable surfaces + sandbox command), `docs/enterprise/README.md` (index), `docs/enterprise/08-ledger-and-postings.md`, `docs/enterprise/43-monitoring.md`, `.env.sample`.

## Tests (new, `__tests__/enterprise/`)
`credit-note-numbering`, `betterstack-telemetry`, `reversal-engine`, `ledger-balance-snapshot`.

## Intentionally deferred
Gateway-side payout transfer reversal (#716 tail), org Novu notification on disputes (#779 §D), broader `recordSystemEvent` adoption.

---

## v1 money-audit — remediation round (this session)

A deep audit of the **money subsystem** (ledger / payments / payouts / refunds / disputes / tax / invoicing, #776) was run against the merged code, and the genuinely-pending correctness, accounting, concurrency, and compliance gaps were closed. Grouped below; `[critical]` items are money-loss / under-payment class.

### Refund engine & reversal (most impactful)
- **A1 `[critical]`** — `OrganizationEarnings.refundedAmountPaise` now tracks the **ORG-SHARE only** (was org+consultant), matching the dispute path and the org-payout `readyAmount` calc. Fixes orgs being under-paid / payout batches blocked on refunded org bookings.
- **A2** — the gateway (Razorpay/Stripe) refund webhook now routes through the canonical **`applyRefundCascade`** instead of the earnings-only legacy path. One engine across **app / cron / webhook** (legs + wallet + earnings + ledger + utilization + GST credit-note reversal).
- **A3/A4** — new `Refund.cascadedAt` + an **atomic claim** at the start of `applyRefundCascade` make the cascade **exactly-once** across all callers — kills the cron↔webhook double-cascade and the wallet-cache double-credit. The backstop cron now selects on `cascadedAt IS NULL`.
- **A6** — refund ratio threaded into `reverseBookingUtilization`, so the `CREDIT_POOL` money-meter / usage ledger reverse **proportionally** to the refund (`REFUND_BOOKING_COHERENCE`), not the ceil'd engagement count.
- **A5** — documented, no code change (the platform-fee rounding residual has no cleaner balanced home and is invariant-safe). **A7** — redundant filter removed.

### Booking ledger completeness & integrity
- **C1** — the booking ledger `DISCOUNT` plug is now derived from the **actual funding-leg debits** (not `payment.amount`), fixing a silently-dropped / imbalanced posting for referral-credit-funded bookings.
- **E1** — multi-collaborator class/webinar bookings now post **per-collaborator `CONSULTANT_PAYABLE`** entries (previously deferred → the journal silently omitted an entire booking class); reconcile's `EARNINGS_LEDGER_DRIFT` now covers them.
- **E2** — the real `Σ(DEBIT) == Σ(CREDIT)` DB-constraint **trigger** now exists (`prisma/sql/ledger-triggers.sql`, applied by `npm run db:triggers`) — the schema previously *claimed* a trigger that didn't exist. **Ops note:** run `npm run db:triggers` after every `prisma db push` / reset (Prisma doesn't manage triggers).
- **F2** — `OrganizationInvoice` total invariant: hard-throw at issue time (`total == subtotal + CGST + SGST + IGST`) + an `INVOICE_TOTAL_MISMATCH` reconcile sweep.

### Concurrency / cron hardening
- **B1 `[critical]`** — abandoned wallet-top-up cleanup re-asserts `status:PENDING` + `providerPaymentId:null` in the `deleteMany`, so a top-up confirmed mid-cleanup (paid) can't be hard-deleted (money loss).
- **B2/B3** — `release-earnings` and `process-payouts` now use **atomic read+claim** (Serializable tx / conditional `updateMany`), plus GitHub Actions `concurrency:` guards — no double-counted releases, no duplicate gateway submits / false `FAILED`.
- **B4** — `reconcile-payment-status` guards its update on the read status, so a racing webhook transition (`PENDING→SUCCEEDED`) isn't clobbered to `EXPIRED`.

### Tax & compliance
- **D1** — deterministic **CGST/SGST split** (`floor` CGST, SGST absorbs the odd paise) so `cgst+sgst == taxPaise` exactly. The old `round`/`round` over-stated odd-tax invoices by 1 paise (which flowed into invoice totals + ledger).
- **D2** — consultant payouts consolidated onto the **Section 194-O** engine (`compliance/tds.ts`); they previously used 194J (10%) — **~100× over-withholding** for an e-commerce operator. Added `TDSRecord.tdsSection` for Form-26Q audit. *(Live payouts are flag-gated OFF, so nothing has moved wrong yet.)*
- **F3** — `indianFiscalYear` reckoned in **IST**, not UTC (invoice + credit-note numbering), so the FY matches the IST issue date (was mis-filing early-April-IST invoices under the prior FY).

### Other
- **F4** — dispute state-machine guard (`lib/payments/dispute-status.ts`) rejects illegal / terminal re-transitions (defends the `LOST` chargeback path against duplicate / out-of-order gateway events).
- **F5** — `recordSystemError` (→ BetterStack) on swallowed booking/refund/clawback ledger postings, so dual-write drift **pages immediately** instead of waiting for nightly reconcile.
- **F1** — verified **NON-ISSUE**: the manual org-payout route's existing race-safe claim pattern already prevents duplicate payouts + orphans; a `@@unique` would wrongly block re-payout after a `FAILED` one. No change.

### PR review (Codex / chatgpt-5.5) resolutions

| Sev | Finding | Status |
| --- | --- | --- |
| P0 | Gateway refunds bypass the canonical cascade | **FIXED** (A2) |
| P0 | Unbilled `INVOICE` refund hits `PaymentLeg @@unique` | **FIXED** — decrements the existing accrual leg instead of inserting a duplicate same-source leg |
| P0 | Org chargeback wallet recovery not idempotent | **ALREADY GUARDED** (`alreadyPosted` ledger-key check) + F4 |
| P1 | Refund-driven credit notes not idempotent | **ALREADY GUARDED** (`CreditNote.refundId @unique` + probe) |
| P1 | CN can be issued against a `DRAFT` invoice | **FIXED** — `mintRefundCreditNote` gates on `ISSUED` / `issuedAt` |
| P1 | Org-invoice-payment refund mints no CN | **FIXED** — new `mintInvoiceRefundCreditNote` |
| P1 | Gateway CN failures swallowed, no durable retry | **FIXED** — cascade error rethrows; tx rolls back (`cascadedAt` reverts) so gateway/cron retry |
| P1 | Reversal engine / `CLASS_MULTI` is only test-wired | **DEFERRED** (tracked) — unfinished production wiring for consolidated-class refunds, not a regression |

### Code-quality
Convertible raw SQL was moved to the **Prisma ORM**: `walletDebit`/`walletCredit` atomic increment/decrement guards, seat-count, invoice/credit-note counters (via `upsert`), top-up-refund. A few raw queries **remain by necessity** — each annotated in code:
- signed-sum `GROUP BY` with a join (org analytics);
- `unnest(languages)` Postgres array (explore-experts);
- `SELECT 1` liveness probes; the ledger trigger DDL.

### Verification
- `tsc --noEmit` **clean**; **46 enterprise/payments suites / 337 tests** pass (mocks/fixtures updated to the ORM + new behavior; added a GST-split test and a DRAFT-credit-note test).
- **Recommended pre-merge gate:** `prisma db push --force-reset` → seed → `npm run db:triggers` → `scripts/reconcile/reconcile-ledgers.ts` (expect `ok:true`; now also covers multi-collaborator bookings + invoice totals) → `scripts/smoke/org-payout-sandbox-smoke.ts` (live-payout gate still holds).

## Verification & live battle-test

Beyond the static gate above, the money core + enterprise API surface were exercised live against a freshly-seeded DB. Coverage is reported honestly — see the "not individually hit" list.

**Methodology**
- [x] Reset + reseeded the dev DB (`prisma db push --force-reset` + `SEED_MODE=small` seed) — 4-org cohort (Wipro SPONSOR/INVOICE, LearnPro HOST, IIT HYBRID/WALLET, solo).
- [x] Applied the ledger balance trigger (`npm run db:triggers`) and validated it (`scripts/db/validate-ledger-trigger.ts`): allows a balanced post, rejects an unbalanced one at COMMIT.
- [x] Built a production bundle (`next build` + `next start`) — the dev server is unstable in this env (Node 26 + Tailwind, unrelated to app code), so testing ran against the prod server.
- [x] Auth/authz temporarily disabled via an env-gated bypass in `requireApiAuth` (synthetic platform-admin; `requireOrgAccess` grants ADMINs an OWNER stub on any org), then **removed, rebuilt from clean source, and the rails verified restored** (see below).

**Ledger / money correctness (ground truth)**
- [x] `reconcile-ledgers` = **Clean, 0 findings** on the fresh seed, and **again after the full battle test** (creates, updates, deletes, invoice issuance + pay-order, refund, wallet confirm). Covers all invariants incl. the new INVOICE_TOTAL_MISMATCH and multi-collaborator EARNINGS_LEDGER_DRIFT.
- [x] Wallet top-up confirm (direct): balance += via the ORM `increment`, balanced TOPUP ledger txn through the trigger, re-confirm idempotent.
- [x] Refund cascade (direct): `cascadedAt` set, earning → REFUNDED, **re-running the cascade is a no-op** (A3/A4 exactly-once proven on real data).
- [x] DB state verified via Supabase MCP: issued invoices `IIT-MADRAS-2026-0001/0002` both `total == subtotal + cgst + sgst + igst` (F2 holds), 18% IGST correct, invoice counter gapless at `nextSeq 3`.

**API route coverage (honest)**
- Scope: **70 enterprise route files / ~120 method-endpoints** (66 under `organizations/**`, 2 admin, 2 overage).
- [x] **All 70 route files exercised** (every route path received ≥1 request).
- [x] **~90 of ~120 method-endpoints hit**, **zero 500s**. Every non-2xx was correct backend logic — validation 400s, cross-org/capability 404s, the >5-member anti-spam 403, intentional 405/409.
  - GET: ~60 endpoints incl. every `[id]` sub-route (on IIT/HYBRID), plus 37 GET routes × Wipro/LearnPro/IIT earlier.
  - Writes (full CRUD lifecycles): rate-card (create→delete), SCIM token (create→delete), purchase order (create→PATCH→delete), contract→program→assignment chain (create→PATCH→delete), invoice ×2 (create+issue→pay), consent (grant→withdraw), outbound webhook (create→GET→rotate-secret→deliveries→delete), member-invite (correct 403), domain-claim (create), data-export (request), SSO settings (PATCH) + provider (validation), payout batch (correct 409), admin-verify (validation), members-bulk (intentional 405).
- Not individually hit (~25–30 endpoints — lower-risk, unchanged by this PR; GET paths covered, write paths follow the same validated handler pattern): some `[id]` PATCH/DELETE (`members/[memberId]`, `invitations/[invitationId]` DELETE, `payouts/[payoutId]` PATCH, `contracts/[contractId]` DELETE, `rate-cards/[cardId]` PATCH, `sso/providers/[providerId]` DELETE, `domain-claims/[domain]` DELETE + `/verify`), `branding/[asset]` upload/delete, `webhooks/.../redeliver`, `overage/[overageEventId]/order`, org-create POST, `invitations/accept`, `scim/group-mappings` POST/DELETE, `data-exports/[id]/download`.

**Auth rails restored (post-test)**
- [x] Bypass removed from source (`grep TEST-AUTH-BYPASS` = 0; `lib/auth-helpers.ts` back to original, no diff), clean rebuild.
- [x] Verified on the bypass-free server: unauthenticated → **401**, real org owner → **200**, owner hitting a non-member org → **403** (authn + authz both enforced).

**Static checks**
- [x] `tsc --noEmit` clean; `jest __tests__/enterprise __tests__/payments` = **46 suites / 337 tests** pass.

### Tracked items (all resolved in later rounds except #780)
- ~~**B5** stuck-webhook sweeper~~ — ✅ resolved in Round 4 (`26772ea9`): active re-driver replaces the passive 5-min staleness window.
- ~~**B6** `INVOICE` credit-limit TOCTOU~~ — ✅ resolved in Round 4 (`ab241b0a`): exposure re-checked inside the Serializable booking tx.
- ~~**C3** overage `OVERAGE_INVOICE_ACCRUAL` leg vs leg-sum invariant~~ — ✅ resolved (`7f7e7d12`): carve `basePaise` out of the base leg; live-verified Σlegs == amount, reconcile Clean.
- The **Int32→BigInt** money widening (**#780**) is tracked separately and intentionally out of scope.

## Round 2 — end-to-end flow testing + new fixes (overage, TDS, payouts)

After the static + battle-test round above, I drove the **overage settlement** and **org/consultant payout** flows end-to-end against the freshly-seeded DB. That surfaced two real money bugs (one critical, one high) plus a latent symmetric defect, all fixed in `7f7e7d12`. Each fix was re-verified live with `reconcile-ledgers` Clean throughout.

### C3 overage leg-sum (CHARGE_ORG) — critical, FIXED
The `CHARGE_ORG` overage path added an `OVERAGE_INVOICE_ACCRUAL` leg of `marginalPaise` **on top of** a full-price base `INVOICE_ACCRUAL` leg. Because the invoice rollup sums **both** leg sources, the org was double-billed by `basePaise` and `Σlegs > amount` (the `PAYMENT_LEG_SUM_MISMATCH` invariant fired). Per the code's own model — `coveredPaise + basePaise == price`, where only the surcharge is genuinely-additional money — the fix carves `basePaise` **out** of the base leg and bumps `amount` by `marginal − carved`.
- **Verified live:** `Σlegs == amount`, the org is billed the booking price (not 2×), and `reconcile-ledgers` is Clean.

### CHARGE_MEMBER parent carve — latent, FIXED proactively
The same root cause exists on the `CHARGE_MEMBER` branch: the parent's org-funded base leg wasn't carved, so the org is double-billed `basePaise` while the member side-charge also collects it. I traced the checkout to confirm the parent is org-funded for the **full price** (the funding source comes from the billing account, independent of `overageBehavior`). No `CHARGE_MEMBER` program is configured in the seed, so there is **zero current impact** — but the symmetric carve makes the path correct for whenever one is added.

### TDS PAN-on-file over-withholding — high, FIXED
Both payout services passed the **encrypted-PAN ciphertext** as `panNumber: "ENCRYPTED"` to `computeTdsForPayout`. `isValidPan("ENCRYPTED")` is false, so the code took the 194-O **5% no-PAN fallback** instead of the **0.1% with-PAN rate — a 50× over-withholding** when a PAN is actually on file. The fix adds a `panOnFile` signal to `TdsConsultantInput`; both `org-payout-service` and `payout-service` now pass `panNumber: null, panOnFile: !!panEncrypted`. The live-payout gate is OFF, so nothing was mis-withheld in production.
- **Verified live:** a ₹40,000 org payout now withholds **₹40 (0.1%)**, not ₹2,000 (5%).

### Payout service exercised end-to-end
Drove the payout service through a full cycle and confirmed:
- Eligibility math: `readyAmount = ΣorgShare − Σrefunds`.
- TDS computed via the 194-O engine.
- Idempotency: same idempotency key → reused, **no duplicate** payout created.
- Guards: net ≤ 0 → **409**, unverified payout account → **409**, empty window → **409**.

### Tests added
- `__tests__/enterprise/overage-settlement-legsum.test.ts` — **5 cases**: CHARGE_ORG no-surcharge / surcharge / priceCap, and CHARGE_MEMBER full + partial carve.
- Extended `__tests__/enterprise/tds-org-payout-input.test.ts` to exercise the **real** `panOnFile` input shape. The prior test passed a valid plaintext PAN and never the actual `"ENCRYPTED"` ciphertext, which is exactly why it gave false confidence and hid the bug.

### Regression
`tsc` clean; **47 enterprise + payments suites / 344 tests** pass; `reconcile-ledgers` Clean (**0 findings**) throughout every live probe.

---

## Round 3 — end-to-end flow testing & fixes

This round took the round-2 fixes (commit `7f7e7d12`) and drove the **overage settlement** and **org/consultant payout** flows fully end-to-end against a freshly-seeded DB, plus re-ran the adversarial refund + pure-function tax/dispute matrices. `reconcile-ledgers` was **Clean (0 findings)** throughout, `tsc` is clean, and **47 suites / 344 tests** pass.

### New fixes (commit `7f7e7d12`)

| # | Sev | Finding | Fix |
| --- | --- | --- | --- |
| 1 | **critical** | **Overage `CHARGE_ORG` (C3) — double-bill + leg-sum break.** The `OVERAGE_INVOICE_ACCRUAL` leg was added **on top of** a full-price base `INVOICE_ACCRUAL` leg. The invoice rollup sums **both** leg sources, so a fully-over-cap engagement billed the org `price + marginal` (**2×**) and `Σ(legs) ≠ Payment.amount` (`PAYMENT_LEG_SUM_MISMATCH`). | `lib/payments/billing/overage-settlement.ts` — carve `basePaise` **out** of the base leg into the overage leg (per the code's own model `coveredPaise + basePaise == price`; only `surchargePaise` is genuinely-additional money). **Verified live:** `Σlegs == amount`, org billed the price (not 2×), reconcile Clean. |
| 2 | latent | **Overage `CHARGE_MEMBER` — symmetric carve.** The org-funded parent kept the full price while the member side-charge also held the marginal → `basePaise` double-collected. | Same carve applied to the parent. **Latent:** no `CHARGE_MEMBER` program in seed (only Wipro=`CHARGE_ORG`, IIT=`BLOCK`) → zero current impact; path is now correct for whenever one is added. |
| 3 | high | **TDS PAN-on-file over-withholding.** Both payout services passed the ciphertext `"ENCRYPTED"` as `panNumber` → `isValidPan()` false → 194-O **5% no-PAN fallback** instead of **0.1%** = **50× over-withholding** when a PAN is on file. | New `panOnFile: boolean` on `TdsConsultantInput` (an encrypted PAN can't be format-checked); `org-payout-service.ts` + `payout-service.ts` now pass `panNumber: null, panOnFile: !!panEncrypted`. Live-payout gate is OFF → no prod impact. **Verified live:** org payout TDS **4,000 (0.1%)**, not 200,000 (5%). |
| 4 | — | **Tests.** | New `__tests__/enterprise/overage-settlement-legsum.test.ts` (**5 cases** — `CHARGE_ORG` + `CHARGE_MEMBER` leg-sum invariant); closed a false-confidence gap in `tds-org-payout-input.test.ts` (it was passing a valid plaintext PAN, never the real `"ENCRYPTED"` input the service constructs). |

### Testing performed this round (all green)

- [x] **Overage end-to-end** — drove `recordOverageAtCheckout` on real over-cap scenarios (`CHARGE_ORG` + traced `CHARGE_MEMBER` funding via `checkout.ts` line 2392 / 1817); confirmed leg-sum invariant + invoice non-double-bill.
- [x] **Org-payout service end-to-end** — eligibility (`readyAmount = ΣorgShare − Σrefunds`), TDS 194-O 0.1%, idempotency (same key → reused), and all guards (net ≤ 0 → **409**, unverified account → **409**, empty window → **409**).
- [x] **Refund adversarial** — over-refund / over-refund-remainder / double-refund all rejected; `cascadedAt` idempotency (re-cascade is a no-op).
- [x] **TDS / GST / dispute pure-function matrix** — 194-O 0.1% / 5%, inter-state IGST, dispute F4 transition guard.
- [x] **`reconcile-ledgers` Clean (0 findings)** throughout; `tsc` clean; **47 suites / 344 tests** pass.

---

## Round 4 — zombie-state recovery (stuck/deadlock states)

A multi-agent **zombie-state audit** mapped every stuck / deadlock / zombie state and recovery gap across **11 money subsystems** (**101 findings → 16 scenarios + 15 recovery gaps**). Diagnosis: defenses are strong at the **transaction boundary**, but recovery was **PASSIVE** — it relied on payment-gateway webhook redelivery plus read-only reconcilers that **detect** drift without **repairing** it. The top findings were then **adversarially verified against the real code**: **8 confirmed real**, **2 refuted as phantoms** (`HELD`-earnings-no-cron is handled inline at dispute resolution; `PAID`-orphan-earnings runs in the same Serializable tx → atomic). The theme of the round: turn **PASSIVE** recovery into **ACTIVE** re-drivers.

### Fixes shipped

| # | Commit | Sev | Finding & fix |
| --- | --- | --- | --- |
| 1 | `26772ea9` | **critical** | **B5 stuck-webhook sweeper.** Webhook routes return HTTP 200 **before** the `after()` money side-effects run, so a crash mid-callback strands `WebhookEvent processed=false` with no redelivery (→ PAID-but-ISSUED invoices, frozen ACCRUED overages, uncredited top-ups, frozen tentative appointments). Extracted the Razorpay dispatch into a **Next-agnostic module** shared by the route + a new **active re-driver** (`scripts/cleanup/sweep-stuck-webhook-events.ts`, 10-min cron) that replays stuck events through the same idempotent handlers. **Live-verified:** a captured-but-uncredited top-up recovers; idempotent; reconcile Clean. |
| 2 | `26772ea9` | **critical** | **org-payout flag-off zombie.** `processOrgPayout` claimed `PENDING→PROCESSING` even with `ENABLE_LIVE_PAYOUTS` off (no gateway/webhook → stuck `PROCESSING` forever). Now stays `PENDING` until live. |
| 3 | `26772ea9` | high | **refund-cascade share cap.** `refundedShareAmount` could overshoot `consultantSharePaise` on a 2nd reversal; capped via `Math.min`. |
| 4 | `738d8130` | high | **refund×chargeback netting.** An app refund + a lost chargeback on the **same** org-funded payment debited the org **twice** (separate idempotency keys). Per the chosen design, a shared **payment-level guard** nets the 2nd reversal against the 1st (`applyOrgChargeback` settles only the un-reversed remainder; `refundPayment`'s refundable subtracts prior `LOST`/`CHARGE_REFUNDED` disputes) → one disputed payment debits the org **exactly once**. |
| 5 | `ca6e9073` | **critical** | **captured-but-reaped top-up recovery.** A ledger-post failure inside `confirmTopUp` (one `$transaction`) rolled the `CONFIRMED` claim back to `PENDING`, then the 6h cleanup **hard-deleted** the gateway-captured row — losing the org's money. Added `WalletTopUp.capturedAt` (stamped **outside** the tx so it survives the rollback), a cleanup guard (`capturedAt IS NULL`), and a 30-min reconciler that re-runs the idempotent confirm. **Live-verified.** |
| 6 | `8a924d41` | **critical** | **false-FAILED payout double-pay.** `process-payouts` marked `PROCESSING→FAILED` + unlinked earnings even when the gateway **already disbursed** (post-submit DB write failed), so a re-batch under a fresh `idempotencyKey` = double disbursement. Catch now distinguishes **pre-gateway** failure (FAIL + release) from **post-gateway** DB failure (persist gateway id, leave `PROCESSING`, keep earnings `LINKED` so `create-payout-batch` can't re-batch them; `handle-stuck-payouts` reconciles vs the gateway). Gate is **OFF** — no prod impact. |
| 7 | `59f6e038` | high | **circuit-breaker freed of abandoned overage charges.** Never-paid `PENDING CHARGE_MEMBER` side-charges counted toward `maxOveragePerCyclePaise` (hard-blocking legit bookings); no `payment.failed` webhook frees an abandon-before-pay. New **daily sweeper** FAILs `PENDING CHARGE_MEMBER` `OverageEvent`s >7d old whose side-`Payment` never `SUCCEEDED` (member can still retry `FAILED→PENDING`). **Live-verified:** ceiling freed. |
| 8 | `ab241b0a` | high | **B6 credit-limit TOCTOU.** The `INVOICE` exposure aggregate ran on the **global client** before the per-slot lock + Serializable tx, so two concurrent disjoint-slot bookings straddled the ceiling. Now **re-checked inside the Serializable tx** (SSI aborts the racing loser); pre-lock check kept as fast-fail. Mirrors the proven `maxUses` re-validation. |

### New recovery infrastructure

4 new sweepers / reconcilers (`jobs/` + `app/api/cleanup/` + GitHub Actions crons):

- `sweep-stuck-webhook-events` — **10-min** cron
- `sweep-orphaned-topup-captures` — **30-min** cron
- `sweep-abandoned-overage-charges` — **daily** cron
- `reconcile-failed-payouts` logic **folded into** the existing `handle-stuck-payouts`

### Gate (all green)

- `tsc` **clean**
- **51 suites / 358 tests** pass (each fix has a regression test)
- `reconcile-ledgers` **Clean** throughout
- every fix **live-verified** where it touches data


teetangh:
chatgpt 5.5 codex feedback

**[P0] Gateway refunds still bypass the new canonical refund cascade**

Target: `app/api/webhooks/utils.ts` around `handleRefundCreated` / `runRefundSideEffects`.

Real Razorpay and Stripe refund webhooks still run the legacy split path: `refundEarnings`, `reverseBookingUtilization`, and referral restoration directly. They never call the new `applyRefundCascade` / `applyReversal` path, so successful gateway refunds miss the v1 additions in `lib/payments/operations/refund.ts`: credit-note creation, invoice-leg reversal semantics, org-earnings clawback handling, wallet/refund ledger discipline, and the new coherence assumptions.

This is especially risky because the cleanup cron only looks for SUCCEEDED refunds whose downstream earnings have not been cascaded yet. Since the webhook path already mutates earnings, the cron can skip the refund and the new canonical cascade never catches up.

Please route successful payment refunds through one canonical path exactly once, preferably `applyRefundCascade` or `applyReversal({ kind: "BOOKING" })`, and remove the duplicated legacy side effects from the webhook branch. Add a webhook-level regression test for an org-funded invoiced payment refund that proves the canonical cascade runs and a credit note is emitted once.

teetangh:
chatgpt 5.5 codex feedback

**[P0] Unbilled INVOICE refunds can hit the `PaymentLeg` uniqueness constraint before the new credit-note path runs**

Target: `lib/payments/operations/refund.ts` around the `INVOICE_ACCRUAL` / `OVERAGE_INVOICE_ACCRUAL` branch; related schema constraint: `PaymentLeg @@unique([paymentId, source])`.

For unbilled invoice-accrual refunds, this branch appends a negative sibling `PaymentLeg` using the same `paymentId` and the same `source` as the original positive leg. The schema only permits one row per source per payment, so this path can throw P2002 for any payment that already has an `INVOICE_ACCRUAL` or `OVERAGE_INVOICE_ACCRUAL` leg. That means the refund cascade fails before the newly added credit-note creation, audit row, and ledger reversal can complete.

Please either model invoice adjustments with a distinct source/model, relax and update the leg uniqueness invariant with all readers adjusted, or update the pending rollup without inserting a duplicate source row. Add a regression test for an unbilled invoice-accrual refund that proves it does not violate `@@unique([paymentId, source])` and still produces the expected correction.

teetangh:
chatgpt 5.5 codex feedback

**[P0] Org chargeback wallet recovery is not idempotent on duplicate terminal dispute events**

Target: `app/api/webhooks/utils.ts`, `applyOrgChargeback`.

The helper debits the org wallet with `walletDebit` before it calls the idempotent `postLedgerTxn` keyed as `chargeback:<disputeId>`. If the gateway sends more than one terminal event for the same dispute, for example `payment.dispute.lost` and then `payment.dispute.closed`, the second invocation will decrement `BillingAccount.walletBalance` again, while `postLedgerTxn` returns `created: false` and does not add a second ledger entry. That creates immediate wallet cache drift and over-recovers from the org.

Please claim/check the chargeback idempotency key before mutating the wallet cache, or only run `walletDebit` when the ledger claim is newly created. Add a duplicate-dispute regression test that invokes the same lost/closed dispute path twice and asserts wallet balance, ledger balance, and audit effects happen exactly once.

teetangh:
chatgpt 5.5 codex feedback

**[P1] Refund-driven credit notes are not idempotent**

Target: `lib/payments/operations/refund.ts`, the new `generateOrgCreditNoteNumber` / `tx.creditNote.create` block.

Every retry of `applyRefundCascade` for the same successful refund allocates a fresh credit-note number and inserts another `CreditNote`. `CreditNote.refundId` is just a nullable string with no uniqueness constraint, and the code does not check for an existing note before allocating the next sequence. A webhook redelivery, cleanup retry, or manual re-run can therefore create duplicate GST credit notes for one refund and permanently consume extra gapless sequence numbers.

Please make refund-driven credit note issuance idempotent, likely by enforcing uniqueness for non-null `refundId` or by upserting/probing before sequence allocation. Add a retry test that runs the same refund cascade twice and asserts exactly one credit note is present and no second credit-note number is allocated.

teetangh:
chatgpt 5.5 codex feedback

**[P1] The new reversal engine and `CLASS_MULTI` refund path are only test-wired**

Target: `lib/payments/operations/reversal-engine.ts`.

The PR says the unified reversal engine fixes multi-booking CLASS refunds, but repo search shows `applyReversal` is only called by `__tests__/enterprise/reversal-engine.test.ts`. Production refund webhooks, cleanup jobs, and app-initiated refunds still call `applyRefundCascade` or older bespoke paths directly, and no production code resolves a consolidated class purchase into child payment ids and calls `applyReversal({ kind: "CLASS_MULTI" })`.

So the new `CLASS_MULTI` logic is currently dead code from a product perspective: tests prove the helper can split child payments, but no real refund path can reach it. Please add a production caller for consolidated class refunds, or narrow the PR claim/checklist until that wiring exists. The fix should include an integration test from the real class refund entry point through `applyReversal`.

teetangh:
chatgpt 5.5 codex feedback

Review scope note for the fixing agent: I focused this pass on v1 backend/money behavior claimed by this stacked PR, not deferred v2/v3/v4 roadmap items. I intentionally did not flag known later-audit gaps such as onboarding UX, dunning/action center, or the broader Int-to-BigInt migration.

Local verification before leaving the comments:

- `npx prisma validate` passed.
- `npx tsc --noEmit` passed.
- `npx jest __tests__/enterprise/reversal-engine.test.ts __tests__/payments/refund-operation.test.ts --runInBand` passed.

The comments above are behavioral/idempotency issues that are not caught by those checks.

teetangh:
chatgpt 5.5 codex feedback

**[P1] Org-invoice payment refunds still mark the invoice REFUNDED without minting a CreditNote**

The new `mintRefundCreditNote` helper covers booking `Payment` refunds, but the org-level invoice-payment refund branch still bypasses it entirely. When `handleRefundCreated` resolves `providerPaymentId` to an `OrganizationInvoice`, it updates the invoice to `REFUNDED`, writes audit, and optionally credits wallet funds, but it never creates a `CreditNote`.

This is the exact post-invoice reduction case: the invoice has already been paid/refunded at the gateway, and CGST Sec 34 is keyed to a tax invoice that has been issued and later needs its taxable value/tax reduced. So the PR can still leave a paid org invoice in `REFUNDED` state with no credit-note artifact for finance/GSTR-1.

Please add an invoice-level CN path, either by generalizing the helper or adding a sibling helper keyed by the gateway refund id, e.g. `refundId: invoice:${gateway}:${refundId}` so partial invoice refunds remain idempotent per gateway refund. Add a webhook regression test that replays the same invoice refund twice and asserts the invoice is `REFUNDED` and exactly one `CreditNote` exists.

teetangh:
chatgpt 5.5 codex feedback

**[P1] `mintRefundCreditNote` can issue a credit note against a DRAFT invoice**

`mintRefundCreditNote` only checks that `Payment.billableToOrgInvoiceId` exists; it does not load or gate on `OrganizationInvoice.status` / `issuedAt`. But `rollupOrgInvoiceAccruals({ issueImmediately: false })` creates a `DRAFT` invoice and still stamps `Payment.billableToOrgInvoiceId`, so a refund before invoice issue can now create an `ISSUED` credit note for an invoice that the app itself treats as legally unissued.

That breaks the lifecycle boundary already documented in the PDF route: DRAFT invoices are not legally issued. CGST Sec 34 also starts from one or more tax invoices having been issued; before that point the right behavior is to adjust/unbill the draft invoice, not create a statutory credit note.

Please select `status`/`issuedAt` in the helper and only mint for issued lifecycle states (`ISSUED`, `OVERDUE`, `PAID`, possibly existing `REFUNDED` replays). For `DRAFT`, return no CN and leave the pending-accrual adjustment to the unbilled-invoice refund fix tracked in #786. Add a unit test proving a DRAFT invoice does not allocate a credit-note number or create a `CreditNote`.

teetangh:
chatgpt 5.5 codex feedback

**[P1] Gateway credit-note failures are swallowed with no durable retry**

The gateway refund path now calls `mintRefundCreditNote`, but it wraps that call in `try/catch` and only logs failures. If `refundEarnings`, utilization reversal, and referral restoration succeed but credit-note creation has a transient DB/sequence error, the transaction still commits the refund side effects with no durable “CN missing” marker.

The existing cascade cron will not reliably repair this: it only selects succeeded refunds whose earnings still look un-cascaded (`refundedShareAmount = 0` plus non-terminal status). A webhook where earnings were already mutated but CN minting failed is therefore likely invisible to the retry job, leaving an invoiced refund permanently missing its tax artifact.

Please either make credit-note minting failure abort the webhook transaction so gateway redelivery retries it idempotently, or add a durable reconciliation path that finds `SUCCEEDED` refunds on invoiced payments with no matching `CreditNote.refundId` and calls the helper. Add a regression that simulates CN failure after the refund side effects and proves the next webhook/reconcile attempt creates exactly one credit note.

teetangh:
chatgpt 5.5 codex feedback

**[P1] Unbilled invoice refunds now avoid P2002 but still trip `PAYMENT_LEG_SUM_MISMATCH`**

The #786 fix changed unbilled `INVOICE_ACCRUAL` / `OVERAGE_INVOICE_ACCRUAL` refunds from “create a negative same-source leg” to “decrement the existing accrual leg in place.” That avoids the `@@unique([paymentId, source])` crash, but the reconciler still enforces `sum(PaymentLeg.amountPaise) === Payment.amount` for every org payment.

After a partial unbilled invoice refund, `Payment.amount` remains the original payment amount while the invoice-accrual leg is reduced, so `scripts/reconcile/reconcile-ledgers.ts` will now report `PAYMENT_LEG_SUM_MISMATCH` on the exact path this PR fixed. This was called out in #786 as the key ripple effect of choosing the in-place decrement fix.

Please either update the invariant to understand refunded/unbilled accrual corrections, or model the correction separately so the original leg-sum invariant remains true. Add a regression that performs an unbilled invoice-accrual refund and then proves the rollup amount is corrected and the ledger reconciler emits zero `PAYMENT_LEG_SUM_MISMATCH` findings.

teetangh:
chatgpt 5.5 codex feedback

**[P1] Refund ledger posting failures still mark the refund cascaded with no durable retry**

`applyRefundCascade` now claims idempotency by setting `Refund.cascadedAt` at the start, but the `REFUND` ledger posting is still inside a `try/catch` that logs and continues. If `postLedgerTxn` fails, the transaction commits wallet/earnings/utilization/CN side effects and leaves `cascadedAt` set, so `cascade-refund-earnings` will never retry it.

The current reconciler does not appear to detect “SUCCEEDED/cascaded refund missing `LedgerTransaction(idempotencyKey = refund:<refundId>)`.” `LEDGER_TXN_IMBALANCE` only checks existing transactions, snapshot drift only checks posted entries, and `EARNINGS_LEDGER_DRIFT` compares original booking postings, not refund counter-postings.

Please make refund ledger posting a required part of a successful cascade, or add a durable retry/reconcile path for missing refund ledger transactions. A good regression is: force `postLedgerTxn` to fail during `applyRefundCascade`, assert `cascadedAt` does not permanently suppress retry, then rerun and assert exactly one `REFUND` ledger transaction exists.

teetangh:
chatgpt 5.5 codex feedback

**[P0/P1] Payout processing has two live processors with different claim and idempotency semantics**

The scheduled GitHub job imports `scripts/payouts/process-payouts.ts`, which now conditionally claims `APPROVED -> PROCESSING` before submitting to the gateway. But the admin and cleanup API routes import `processApprovedPayouts` from `lib/payments/payouts`, whose `processSinglePayout` still does an unconditional status update before gateway submission.

Those two processors also appear to use different gateway idempotency keys: the script uses `payout_${payout.id}`, while the service can use `ConsultantPayout.idempotencyKey` (`payout_${consultantProfileId}_${batchId}`). If a manual/API trigger overlaps the scheduled job, both can submit the same payout with different gateway idempotency keys.

Please collapse to one payout processor or make every caller use the same atomic claim and the same gateway idempotency key. Add a race regression that simulates the scheduled job and API path reading the same APPROVED payout and proves only one gateway submission is attempted.

teetangh:
chatgpt 5.5 codex feedback

**[P1] Ledger balance trigger is not Prisma-managed, so v1 needs an explicit deploy gate**

The new DB-level double-entry backstop is valuable, but `ledger-triggers.sql` is applied only by `npm run db:triggers`; Prisma migrate/db-push will not install it. Unless the v1 reset/deploy path runs that script, production can still accept unbalanced manual/future raw writes despite the PR claiming DB-level enforcement.

Please wire this into the documented deployment/reset sequence and CI smoke path, or add a checked script that fails if the trigger is missing. The acceptance test should prove an unbalanced `LedgerEntry` set cannot commit after a fresh reset/push.

Issue #807, opened 2026-06-05
GST invoice/credit-note numbers exceed CGST Rule 46(b)/53 16-character cap


teetangh:
Found in the #789 enterprise audit (CGST research bundle) and verified in code.

\`lib/payments/billing/{invoice,credit-note}-numbering.ts\` concatenate the org prefix without bound. CGST Rule 46(b) (invoices) and Rule 53 (credit notes) cap a document number at **16 characters**. The credit-note format \`<PREFIX>-CN-<FY>-<SEQ>\` overflows for any prefix longer than three characters:

- \`WIPRO-CN-2026-0001\` = 18 chars (breach)
- \`ACME-CN-2026-0042\` = 17 chars (breach)
- slug fallback \`ACME-CORP-CN-2026-0001\` = 22 chars (gross breach)

Several existing tests asserted these non-compliant numbers, confirming the defect was baked in. Fix caps the prefix per-format so every number fits 16 chars (and shrinks the budget automatically if a sequence ever exceeds four digits). Pre-MVP, so the format change is safe; orgs with long slugs should configure a short \`invoiceNumberPrefix\` for cleaner numbers.

Fix + before/after regression tests in the PR. Part of #789.

Issue #806, opened 2026-06-05
Razorpay dispatch silently drops payout.failed, dispute.under_review, and dispute.action_required


teetangh:
Found in the #789 enterprise audit and verified in code.

The dispatch switch in \`app/api/webhooks/razorpay-dispatch.ts\` routes only a subset of events; three terminal/transition events fall through to the \`default\` "Unhandled" branch even though their handlers are already written:

- **\`payout.failed\`** — \`handleRazorpayPayoutWebhook\` + \`markOrgPayoutFailed\` handle it (the docstring even says "Called when the gateway sends payout.failed or payout.rejected"), and a unit test calls the handler directly, but the production switch never routes it. A failed org payout is left stuck in \`PROCESSING\` with its earnings never released back to \`READY\`.
- **\`payment.dispute.under_review\`** — never advances \`Dispute.status\` to \`UNDER_REVIEW\`.
- **\`payment.dispute.action_required\`** — the deadline-bearing signal is dropped entirely.

\`mapDisputeStatus\` already maps \`under_review\`→\`UNDER_REVIEW\` and \`action_required\`→\`NEEDS_RESPONSE\` correctly; only the dispatch routing was missing.

Fix + before/after regression test in the PR. Part of #789.

Issue #789, opened 2026-06-05
docs(enterprise): full doc-set rewrite — SDE leveling, ADR band, payments absorption, regulatory refresh


teetangh:
Umbrella for the in-place rewrite of \`docs/enterprise/\` (50 files / ~16.8k lines / 88 Mermaid diagrams), planned 2026-06-05.

## Goals
1. **Readability** — full-English-sentence prose tree-wide; no fragment headers, fragment bullets, or lead-in-less tables.
2. **SDE1→SDE4 leveling** — YAML frontmatter (\`audience\`/\`status\`/\`last-reviewed\`) on every doc + four curated reading paths in the README.
3. **New \`70-design-decisions/\` band** — 12 ADRs grounding the why (double-entry #772, integer paise, PENDING_TRUST #687, etc.).
4. **Payments absorption** — org-side payouts/refunds/disputes/webhooks move from \`docs/payments/\` into \`10-money-and-ledger/\` (renumbered to 13 files); \`docs/payments/\` keeps B2C with redirect callouts.
5. **Full independent regulatory refresh** — CGST Rules 46/53, IT Act 2025 §393, DPDP Act+Rules, MSME §15/§43B(h), RBI PA, CPA 2019, Razorpay/RazorpayX/IRP mechanics, from primary sources, with explicit divergence flags vs \`docs/compliance/\`.
6. **Diagram completeness** — state diagram per status enum, sequence diagram per money flow, ER diagram per domain, context diagram per band.

\`docs/compliance/\` and \`docs/booking/\` are out of scope (cross-linked only).

Sub-issues track one PR each; PRs land on \`feature/enterprise\`.

PR #808, opened 2026-06-05
fix(enterprise): route dropped webhook events + cap GST document numbers

teetangh:
Stacked on top of the #789 docs series (P8). Ships the two clearly-correct, low-risk bugs surfaced by the audit — each with a **before/after regression test** — and refreshes three stale comments. The remaining audit findings need a design/money-logic call and are listed below for triage rather than fixed here.

### 1. Webhook dispatch drops (#806)
`razorpay-dispatch.ts` routed only a subset of events. Added `payout.failed`, `payment.dispute.under_review`, and `payment.dispute.action_required` cases. The handlers already existed — `payout.failed`'s reconciler, `markOrgPayoutFailed`, and a direct-call unit test were all written, but the switch never reached them, so a failed org payout stayed stuck in `PROCESSING` with earnings unreleased.

**Before:** `__tests__/enterprise/webhook-dispatch-gaps.test.ts` shows all three events logging `📄 Unhandled Razorpay event type` with **0 handler calls**.
**After:** 3/3 route to their handlers.

### 2. GST document-number overflow (#807)
`<PREFIX>-CN-<FY>-<SEQ>` overflowed CGST Rule 46(b)/53's 16-char cap for any prefix > 3 chars (`WIPRO-CN-2026-0001` = 18; slug fallback `ACME-CORP-CN-2026-0001` = 22). A shared `fitPrefixToRule46` helper caps the prefix per format and shrinks the budget if a sequence exceeds four digits.

**Before:** five existing tests *asserted* the over-length numbers — the bug was baked into the suite.
**After:** `gst-doc-number-length.test.ts` (new) plus the corrected existing tests prove every number is ≤16 chars across prefix lengths.

### 3. Stale comments
DPDP operational date (13 not 14 May 2027); MSME deadline citation (Income-tax §43B(h)→§37(2)(g) under the 2025 Act, not "CGST"); the `reconcile-disputes` "Razorpay has no dispute API" note (it now does).

## Testing
All before/after demonstration was done with the project's jest + mocked-prisma harness (the same pattern its existing webhook/payout tests use). I did **not** drive the live `/api/dev/mock-webhook` route for these: it writes to the shared remote Supabase DB and bypasses the production dispatch switch where these bugs live, so it can neither exercise them nor be used without polluting shared data. **429 enterprise + payments tests pass; repo-wide `tsc` clean.**

## Deferred for your triage (verified, but each needs a design/money call — not fixed here)
- 🔴 **Org payout `COMPLETED → reversed` is a silent no-op** — needs a `REVERSED` `PayoutStatus` + ledger-reversal logic (clawback). Money path; wants review.
- 🔴 **GST credit note under-credits tax** — `mintRefundCreditNote` treats the accrual leg as tax-inclusive while invoices are built tax-exclusive, under-crediting the full GST (a test fixture encodes the wrong value). Compliance-grade money math.
- 🔴 **CHARGE_MEMBER over-cap `basePaise` collected from no one** when the member abandons the side-charge — the carved base is never re-billed to the org.
- 🟠 **DRAFT invoices claim accruals + burn gapless numbers** — voiding a draft orphans the payments (never re-billed) and leaves a Rule 46 sequence gap.
- 🟠 **Outbound webhook deliveries stuck in `IN_FLIGHT`** are never recovered (no sweeper for that state).
- 🟠 **Refund webhook arriving before `payment.captured`** is acked and permanently dropped (no retry).
- 🟠 **Stuck `processed=false` events older than 72h** are orphaned between the sweeper window and the 90-day archive.
- 🟡 **xflow/lemon-squeezy signature checks** are non-timing-safe and skipped when the header is absent.
- 🟡 **Dispute `closed` gateway status** maps to `NEEDS_RESPONSE` (masked today by the terminal-transition guard) — needs an enum/semantics decision.


teetangh:
Review feedback addressed in `0b0a83db` on `feature/enterprise` (this PR is merged): the suggested fixes were applied there, and the CI build failure (an ESLint `no-fallthrough` introduced by a comment between two case labels) is resolved in the same commit.

Issue #810, opened 2026-06-05
SCIM offboarding leaves a stale session + SCIM token expiresAt is never enforced


teetangh:
Found in the #789 auth audit and verified in code. **Severity: medium.**

Two SCIM auth gaps:

1. **Deprovision/suspend doesn't invalidate the session.** \`lib/scim/operations.ts\` (reprovision \`active=false\`, \`deprovisionScimUser\`) flips Membership to \`SUSPENDED\` but never calls \`bumpUserSessionGeneration\`, while every in-app member mutation does. Hard API gates are unaffected (\`requireOrgAccess\` re-reads status live), but the cached session payload + 5-minute cookie cache keep listing the user as an active member until it lapses — so an IdP offboard doesn't take effect on session-trusting read/UI surfaces immediately.
2. **\`ScimToken.expiresAt\` is never enforced.** The schema comment says \`requireScimAuth\` must refuse an expired token "even if status = ACTIVE", but \`lib/scim/auth.ts\` only checks \`status\`. A token minted with a TTL authenticates indefinitely.

Fix bumps sessionGeneration in the SCIM suspend/deprovision transactions and enforces \`expiresAt\` in \`requireScimAuth\`; before/after tests included. Part of #789.

Issue #809, opened 2026-06-05
Privilege escalation: MAINTAINER can mint/invite an OWNER via POST /members and /invitations


teetangh:
Found in the #789 auth audit and verified in code. **Severity: high (privilege escalation).**

The members PATCH route correctly refuses to assign the OWNER role unless the actor is already an OWNER (\`members/[memberId]/route.ts\`: "A MAINTAINER renaming someone to OWNER would effectively grant themselves extra privileges by proxy"). But the two creation paths forgot the same guard:

- \`POST /api/organizations/[orgId]/members\` is gated at \`MAINTAINER\` yet \`CreateBodySchema.role\` includes \`OWNER\`, and the create branch never checks the actor's rank — the membership is written \`status: ACTIVE, role: OWNER\` directly.
- \`POST /api/organizations/[orgId]/invitations\` has the same hole; the accept route trusts the invitation's stored role with no inviter re-check.

A MAINTAINER (rank 80) can therefore make an accomplice an ACTIVE OWNER — full billing, deletion, ownership-transfer, SSO/SCIM rights — bypassing the documented "MAINTAINER has no security-sensitive surface" design.

Fix mirrors the existing PATCH guard on both POST routes; before/after test included. Part of #789.

PR #811, opened 2026-06-05
fix(enterprise): close OWNER-escalation hole + SCIM offboarding gaps

teetangh:
Stacked on the webhook/GST fix PR (#808). Three verified auth bugs from the #789 audit, each with a before/after test. **434 enterprise + payments tests pass; `tsc` clean.**

## 1. Privilege escalation — MAINTAINER could mint/invite an OWNER (#809, HIGH)
`POST /members` and `POST /invitations` were gated at `MAINTAINER` but their role schemas include `OWNER`, and neither checked the actor's own rank — so a MAINTAINER could direct-add or invite an **ACTIVE OWNER** (full billing, deletion, ownership-transfer, SSO/SCIM) and escalate by proxy. The members **PATCH** route already guards exactly this ("a MAINTAINER renaming someone to OWNER would effectively grant themselves extra privileges"); the two creation paths simply omitted it. Fix mirrors that guard on both POST routes.

`__tests__/enterprise/owner-role-escalation-guard.test.ts`: a MAINTAINER assigning/inviting `role: OWNER` now gets `403 OWNER_ROLE_REQUIRES_OWNER` before any write; an OWNER passes the guard.

## 2. SCIM offboarding leaves a stale session (#810)
`deprovisionScimUser` and reprovision-to-inactive flipped Membership to `SUSPENDED` without bumping `sessionGeneration`, while every in-app member mutation does — so an IdP offboard lingered in the cached session payload until the cookie cache lapsed. (Hard API gates were unaffected — `requireOrgAccess` re-reads status live — but session-trusting read/UI surfaces were stale.) Now bumped in the same transaction.

## 3. SCIM token `expiresAt` never enforced (#810)
The schema comment says `requireScimAuth` must refuse an expired token "even if status = ACTIVE", but only `status` was checked. Now an expired-but-ACTIVE token returns 401. `__tests__/enterprise/scim/auth.test.ts` gains positive + negative expiry coverage.

## Testing
Same jest + mocked-prisma harness the repo uses for these routes (no shared-DB writes). The escalation test drives the real route handlers; the SCIM tests drive `requireScimAuth`.


teetangh:
Review feedback addressed in `0b0a83db` on `feature/enterprise` (this PR is merged): the suggested fixes were applied there, and the CI build failure (an ESLint `no-fallthrough` introduced by a comment between two case labels) is resolved in the same commit.

Issue #784, opened 2026-05-31
Enterprise — Prisma schema design audit (DRY, redundancy, refactors)


teetangh:
## Why this issue

A software-engineering design audit of `prisma/schema.prisma` — DRY violations, redundant/near-duplicate models, denormalization smells, naming/type inconsistencies, and dead schema. Prompted by the `OrganizationPlan` collapse (which removed one parallel plan shape); this catalogs the rest. Each finding carries a **verified** assessment + a recommendation. The big refactors are documented here rather than executed because they're load-bearing and a pre-launch rewrite is mostly regression risk.

**Already done (committed):** `983e89a4` — dropped the dead three-ledger comment block, the single-value `PayoutArrangement` enum, the write-only `ConsultantProfile.totalRevenue/pendingRevenue` caches (9 writes / 0 reads), and added the missing `Payment.discountCodeId` index. Plus `6fa87df2` collapsed `OrganizationPlan` + its 4 config children into the per-type plans.

**Tracked elsewhere (don't re-open here):** `#780` Int→BigInt money · `#781` currency String→enum + onDelete Restrict + Float→Decimal rates + finance indexes · `#782` overage reconcile.

Tags: 🔴 critical · 🟠 high · 🟡 medium · `(verify)` = assert against code before acting.

---

## 🔴 1. Four parallel plan families — the headline DRY smell

`ConsultationPlan` / `SubscriptionPlan` / `WebinarPlan` / `ClassPlan` each re-declare **~20 identical fields**: `title, description, price, priceCurrency, language, level, prerequisites, materialProvided, learningOutcomes, topics, consultantProfileId, organizationId, visibility, materials, createdAt, updatedAt`, plus the same `@@index([organizationId])` and `@@index([visibility, organizationId])`. The duplication repeats one level down — `Consultation`/`Subscription`/`Webinar`/`Class` booking wrappers, `WebinarCollaborator`/`ClassCollaborator`, the 4× `Topic` joins.

- **Why it's a smell:** the canonical "shotgun surgery" case — adding a plan field means editing four models + four route families + four UIs.
- **Blast radius:** ~400 call sites *each* (plan CRUD, display, booking, pricing, utilization, payout, recording). The most load-bearing models in the system.
[...]

Issue #692, opened 2026-04-20
Referrals + credits + discounts — anti-fraud + correctness audit


teetangh:
# Referrals + credits + discount codes — anti-fraud + correctness audit

> Audit lens: fraud surface + math correctness on the referrals / credits / discount-codes subsystem. Sibling tracker. Does **not** duplicate **#437** (referral qualifying-actions roadmap), **#604** (closed — invoice PDF credit drift), **#677** (payments-side discount + credit ordering).

## TL;DR

- **REF-1 (P1)** — anti-gaming surface is paper-thin. Self-referral check exists; everything else (duplicate accounts, IP/device velocity, circular referrals, mass signup) is unguarded. With current 80/20 split + ₹500 default referrer reward, the ROI on a VPN-farmed cluster of 10 fake accounts is well over 100%.
- **REF-2 (P1)** — `reverseCreditsForPayment` (`lib/referrals/service.ts:451`) restores expired credits when a refund arrives months after expiry. The credit becomes spendable again. Verified by code path inspection.
- **REF-3 (P1)** — referral qualification window check runs **outside** the conditional `updateMany` guard (`lib/referrals/service.ts:261-270`). Concurrent webhooks on a referral right at the day-30 boundary can leave the row in `REWARDED` instead of `EXPIRED`.
- **REF-4 (P2)** — discount `maxUses` race: non-Serializable check runs at price calculation, real check runs inside the Serializable tx ~80 lines later (after Razorpay order creation). Consultee can hit "max uses reached" *after* the payment intent exists, causing UX whiplash.
- **REF-5 (P2)** — credit-expiry cron has no API endpoint at `app/api/cleanup/expire-credits/route.ts` (other cleanup crons do). Manual invocation only.

## What is working well (don't fix)

- **PaymentLeg ledger discipline** — every credit application writes a `PaymentLeg` row with `source: REFERRAL_CREDIT` (`lib/payments/operations/checkout.ts:406-429`). This is the right pattern; refund reversal can join through `paymentLeg.sourceRef`.
- **Cumulative proportional refund math** — `reverseCreditsForPayment` uses `cumulativeTarget - alreadyRestored` (lines 492-502) instead of per-refund deltas, avoiding rounding drift on multi-refund sequences. Correct design.
- **FIFO credit consumption** — credits ordered by `createdAt ASC` so oldest-first. No double-spend possible within a single checkout.
- **Self-referral check** at `service.ts:195` — present and correct.
[...]

Issue #689, opened 2026-04-20
Stream.io / video / recording subsystem production-readiness audit


teetangh:
# Stream.io / video / recording subsystem audit — master tracker

> Audit of the Stream.io integration layer (video calls, recordings, chat, webhooks, transfer pipeline). Sibling tracker to #676 / #677 / #688. Does not duplicate **#471, #472, #473, #535, #639, #342, #367, #366, #360, #248**.

## TL;DR

- **STR-1 (P0)** — recording access does not check refund status. Refunded users keep permanent recording access via `paymentStatus: "SUCCEEDED"` filter that ignores subsequent SUCCEEDED `Refund` rows.
- **STR-2 (P1)** — recording transfer (Stream S3 → Supabase) has a 5-day window before 14-day expiry, with no exponential backoff and 6-hour cron interval. A persistent failure can run out the clock and lose data.
- **STR-3 (P1)** — sync endpoints don't enforce `RecordingStoragePolicy`. STREAM_ONLY recordings are created with same code path as SUPABASE_PERMANENT.
- **STR-4 (P1)** — no audit trail of join/leave per attendee. `MeetingSession.isRecording` tracks state but no per-participant duration → can't reconcile no-show vs full-session.
- **STR-5 (P2)** — webhook idempotency mostly correct (event log + recording-existence check) but the secondary check is fragile if filenames ever collide.

Cross-refs: **#473** (circuit breaker) is the umbrella for resilience; **#535** (hard-delete soft-deleted Stream users) is adjacent.

---

## STR-1 (P0) — Recording access not revoked on refund

**Files**:
- `app/api/stream/recordings/[recordingId]/route.ts:88-114` (consultee access check)
- `lib/stream/recording-service.ts:381` (`getPaidPlanIds` consultee)
- `lib/stream/recording-service.ts:902` (consultee enrollment lookup)

**Verified**: yes — read all three sites. Each filters payment via `paymentStatus: "SUCCEEDED"` and treats a hit as access. None checks for related `Refund` rows.

**Reproduction**:
1. Consultee enrolls in paid webinar/class.
2. Webinar happens, recording is captured.
3. Consultee requests + receives full refund. `Payment.paymentStatus` stays `SUCCEEDED`; a separate `Refund` row is created with `status: SUCCEEDED`.
4. Consultee fetches `/api/stream/recordings/<id>` → access still granted because the lookup only checks the payment row.

[...]

Issue #440, opened 2026-02-14
Add DB-level exclusion constraint for consultant slot overlap prevention


teetangh:
## Summary

There is no database-level constraint preventing double-booked appointment slots for the same consultant. Currently, overlap prevention relies entirely on application-level checks (distributed Redis locks + Prisma queries), which are vulnerable to race conditions under concurrent writes.

## Problem

When two booking requests arrive simultaneously for the same consultant time slot, it's possible for both to pass the application-level overlap check before either commits, resulting in double-booked slots. The distributed lock mitigates this but doesn't fully eliminate the risk (e.g., lock expiry, network partitions).

## Proposed Solution

Add a PostgreSQL exclusion constraint on the `SlotOfAppointment` table using `btree_gist`:

### 1. Enable the `btree_gist` extension
```sql
CREATE EXTENSION IF NOT EXISTS btree_gist;
```

### 2. Add denormalized `consultantProfileId` column to `SlotOfAppointment`
```sql
ALTER TABLE "SlotOfAppointment"
ADD COLUMN "consultantProfileId" TEXT;
```

Populate from existing data via the appointment → event → plan chain, then add NOT NULL constraint.

### 3. Create exclusion constraint
```sql
ALTER TABLE "SlotOfAppointment"
ADD CONSTRAINT no_consultant_slot_overlap
EXCLUDE USING gist (
  "consultantProfileId" WITH =,
  tstzrange("startsAt", "endsAt") WITH &&
);
```

### 4. Update Prisma schema
Add the `consultantProfileId` field to the `SlotOfAppointment` model and add a relation to `ConsultantProfile`.

### 5. Update application code
- Set `consultantProfileId` when creating `SlotOfAppointment` records
- Handle the exclusion constraint violation error gracefully (Prisma error code P2002 or raw constraint violation)

## Why This Was Deferred

- Requires a Supabase migration + Prisma schema change + data backfill
- The `btree_gist` extension needs to be enabled in Supabase
- Denormalization adds write-time overhead and needs to stay in sync
- Too high-risk to bundle with the booking algorithm bug fix PR

## Related

- Part of booking algorithm bug review (Bug #16 in `bugs/booking-algorithm/16-missing-db-level-overlap-constraint.md`)
- Fixed in PR: application-level overlap checks improved (canonical overlap predicate, consistent lock keys, centralized occupancy policy)

## Labels
enhancement, database, booking-system

PR #876, opened 2026-06-16
fix(stream,referrals,collaborators): recording-access-on-refund + referral correctness + co-host scheduling guard

teetangh:
## Money/access correctness — launch-blockers (batch 1 of 2)

The first of two PRs productionizing the customer-facing subsystems. This one carries the cross-subsystem **P0/P1 money + access correctness** fixes, each with unit tests. The remaining launch-blockers (Documents rate-limit/status-gate/reconcile, Stream circuit-breaker + attendance + recording-expiry email, email DLQ, Better Stack telemetry, and the broader co-host scheduling guard) follow in the next PR.

### STR-1 (P0) — refunded users kept recording access
`PaymentStatus` has no `REFUNDED` value, so a refunded payment stays `SUCCEEDED` and the reversal lives only in `Refund` rows. Every recording-entitlement check filtered on `paymentStatus: SUCCEEDED` alone, so a fully-refunded buyer kept access. New shared helper `lib/payments/utils/refund-balance.ts` (`isPaymentEntitled`) nets refunds against the payment amount and is wired into **all four** entitlement paths:
- `app/api/stream/recordings/[recordingId]/route.ts` (single recording)
- `lib/stream/recording-service.ts` → `getPaidPlanIds` (consultee list + resources API) and `syncRecordingsForConsultee`
- `app/api/stream/meetings/[streamCallId]/recording-info/route.ts`

A full refund revokes access; a partial refund retains it. Chargeback-driven revocation is deliberately deferred.

### REF-2 (P1) — refund restored expired credits
`reverseCreditsForPayment` restored `remainingAmount` onto credits regardless of expiry, resurrecting balance the expiry cron re-zeroes. It now skips (and logs) any credit that expired after it was applied; the usage row stays so the credit reads as consumed.

### REF-3 (P1) — qualification-window TOCTOU
The window was checked in app code (`Date.now()` vs `signedUpAt`) separately from the status guard. It's now folded into the reward `updateMany` WHERE (`signedUpAt >= cutoff`), making reward-vs-expire a single atomic guarded claim against the committed row.

### AE-2 (partial, P1) — co-host double-booking at webinar scheduling
Co-hosts aren't slot participants, so neither the `#440` exclusion constraint nor owner-scoped checks ever saw them — a co-host could be silently double-booked. New `lib/collaborators/availability.ts` (`assertCollaboratorsAvailable`) rejects scheduling a webinar onto a time an accepted co-host is already committed to (→ 409), wired into the webinar `crud-with-plan` PATCH. The robust architectural extension (owner DB-guard on group-event slots + co-host validation across the class/reschedule commit paths) lands in the next PR.

### Verification
- `tsc --noEmit`: 0 errors
- 15 new unit tests (refund-balance, referrals REF-2/REF-3, collaborator availability) — pass
- 219 existing stream + payments tests — pass (no regressions)

Part of #689, #692, #784.

## Summary by CodeRabbit

* **New Features / Improvements**
  * Added refund-aware access checks so fully refunded successful payments no longer grant meeting/recording access.
  * Prevented webinar/class co-host double-bookings during rescheduling; conflicts now return a clear 409 with the clashing co-host name.
  * Improved referral qualification and credit restoration behavior to be transaction-guarded and to respect credit expiration.
* **Tests**
  * Added Jest coverage for collaborator availability conflict detection, refund/entitlement calculations, and referral qualification/credit restoration scenarios.

### DIFF
--- a/__tests__/collaborators/availability.test.ts
+++ b/__tests__/collaborators/availability.test.ts
@@ -28,16 +28,19 @@ function makeDb(opts: {
     },
     slotOfAppointment: {
       findFirst: jest.fn().mockImplementation(({ where }) => {
-        // The per-collaborator profile id lives in the appointment OR filter;
-        // resolve a conflict if that collaborator is in the conflict set.
-        const profileId =
-          where.appointment.OR[0].consultation.consultationPlan
-            .consultantProfileId;
-        return Promise.resolve(
-          opts.conflictForProfileIds?.includes(profileId)
-            ? { id: "conflict-slot" }
-            : null,
+        // The appointment OR filter names a consultantProfileId per commitment
+        // clause. This works for both the batched detection query (every co-host
+        // in one OR) and the per-co-host probe (one co-host's clauses): resolve a
+        // conflict if any named profile is in the conflict set.
+        const orClauses: Array<{
+          consultation?: { consultationPlan?: { consultantProfileId?: string } };
+        }> = where.appointment.OR;
+        const clash = orClauses.some((clause) =>
+          opts.conflictForProfileIds?.includes(
+            clause.consultation?.consultationPlan?.consultantProfileId ?? "",
+          ),
         );
+        return Promise.resolve(clash ? { id: "conflict-slot" } : null);
       }),
     },
   };
--- a/__tests__/collaborators/availability.test.ts
+++ b/__tests__/collaborators/availability.test.ts
@@ -1,0 +1,99 @@
+/**
+ * @jest-environment node
+ */
+
+/**
+ * AE-2 (#784) — co-hosts aren't slot participants, so nothing else catches a
+ * co-host being double-booked when an event is scheduled. These pin the guard:
+ * no-op without collaborators, throw when an accepted co-host overlaps.
+ */
+
+import {
+  assertCollaboratorsAvailable,
+  CollaboratorUnavailableError,
+} from "@/lib/collaborators/availability";
+
+function makeDb(opts: {
+  collaborators: Array<{ name: string | null; consultantProfileId: string }>;
+  conflictForProfileIds?: string[];
+}) {
+  return {
+    collaborator: {
+      findMany: jest.fn().mockResolvedValue(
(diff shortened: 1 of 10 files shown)


teetangh:
Addressed the CodeRabbit + Gemini review findings in 65b5b3fa:

- **availability.ts (CodeRabbit, soft-delete):** the co-host clash scan now filters `deletedAt: null` on both the slot and its appointment, so a cancelled/soft-deleted booking no longer surfaces as a phantom 409.
- **availability.ts (Gemini, N+1):** the per-co-host `findFirst` loop is collapsed into a single batched detection query on the common (no-conflict) scheduling path; the per-co-host probe now runs only in the rare blocking case that needs to name the clashing host.
- **referrals/service.ts (Gemini, optional chaining):** `reverseCreditsForPayment` now optional-chains the `credit` relation.
- **referrals/service.ts (CodeRabbit, nit):** the skipped-restore log now says it reports paise rather than a credit count.

tsc clean; collaborator + referral suites pass.

teetangh:
Addressed the CodeRabbit follow-up review in ee2cd3ce:

- **availability.test.ts (nit):** the mock now extracts the `consultantProfileId` from all six commitment-clause shapes (consultation / subscription / webinar / class owner, plus webinar / class collaborator), so a regression in the collaborator predicates is actually exercised instead of silently passing.
- **referrals/service.ts (Major):** dropped the optional chain on `usage.credit` — it's a required FK relation that the `findMany` always includes, so it can't be null; the `?.` was dead/misleading.
- **availability.ts co-host TOCTOU (Major):** acknowledged and **by design for this PR**. Co-hosts are intentionally not slot participants here; the agreed approach is app-level co-host validation running **inside the Serializable scheduling transaction** (which makes a concurrent same-co-host schedule a serialization conflict in practice), with the fully DB-enforced per-host commitment guard (a dedicated commitment table + exclusion constraint) tracked as a follow-up rather than expanded into this PR. The owner side *is* now DB-enforced via the `slot_no_confirmed_overlap` constraint.

Issue #694, opened 2026-04-20
Document upload / review / storage subsystem audit


teetangh:
# Document upload / review / storage subsystem audit

> Audit of the document subsystem: `AppointmentDocument`, `PlanMaterial`, `ProfileVerificationDocument`, `SupportTicketAttachment`, plus storage layer + reconciliation cron. Sibling tracker. Does **not** duplicate **#346, #347, #348, #676 A8, #688, #689**.

## TL;DR

- **DOC-1 (P1)** — per-download authz doesn't re-check appointment status. A consultee whose appointment is cancelled / refunded keeps reading documents through their previously-cached URL until the 1h signed-URL TTL expires.
- **DOC-2 (P1)** — no upload-rate limiting on document endpoints. `lib/rate-limit.ts` exists but isn't applied to `app/api/appointments/[appointmentId]/documents/**` or `app/api/verification/documents/**`. Storage-quota DoS via 10MB-file flood is open.
- **DOC-3 (P1)** — `reconcile-document-storage` cron detects missing files (DB row exists, Supabase file gone) but **doesn't flag them in the DB**. Always returns `missingFilesMarked: 0`. Users hit 500s on download with no platform-side visibility.
- **DOC-4 (P2)** — `ProfileVerificationDocument` (PAN, Aadhaar, passport scans) relies on Supabase bucket-level encryption only. No application-layer encryption, no `encryptionKeyVersion`, no `virusScanStatus`, no `fileHash`.
- **DOC-5 (P2)** — `Content-Disposition` filename uses `encodeURIComponent(document.originalName)` without further validation. Header-injection-shaped originalNames (newline-bearing strings via decoded `%0a`) aren't categorically blocked.

Cross-refs: **#676 A8** (`reviewedBy` as String not FK — already in booking tracker), **#346** (pagination), **#347** (bulk review), **#348** (real-time updates), **#688 SC-2** (soft-delete on these models).

---

## DOC-1 (P1) — Per-download authz misses status check

**File**: `app/api/appointments/[appointmentId]/documents/[documentId]/download/route.ts:92-117`
**Verified**: yes — read end-to-end. The route confirms (a) the user has access to the appointment, (b) the document belongs to that appointment. It does **not** check `appointment.consultation?.requestStatus` or `appointment.subscription?.requestStatus` against `CANCELLED` / `EXPIRED`.

A user who could access docs five minutes ago can keep reading them after the appointment is cancelled, until either:
[...]

Issue #475, opened 2026-03-08
[RESILIENCE] Sentry error tracking integration


teetangh:
## Problem

All application errors are logged to stdout only. In production (Vercel), these logs are ephemeral and not aggregated. There is no alerting on error spikes, no stack trace grouping, and no user context for debugging.

## Proposed Solution

1. **Global Sentry init** in `instrumentation.ts` (Next.js convention)
2. **Capture unhandled exceptions and promise rejections** — server + client
3. **Attach user context** — userId, role, consultantProfileId/consulteeProfileId
4. **Source maps** for production stack traces
5. **Alert on error rate spikes** — Sentry alert rules
6. **Replace console.error** in critical paths with `Sentry.captureException()`

## Acceptance Criteria

- [ ] Sentry SDK initialized for both server and client
- [ ] Unhandled errors automatically captured with stack traces
- [ ] User context attached to error events
- [ ] Source maps uploaded during build
- [ ] Alert rule for >10 errors/minute
- [ ] `app/error.tsx` reports errors to Sentry

## Implementation Notes

- Use `@sentry/nextjs` package
- `instrumentation.ts` for server-side init
- `sentry.client.config.ts` for client-side init
- Vercel has native Sentry integration for source maps

teetangh:
Verified implemented via PR #901 - @sentry/nextjs, instrumentation.ts, instrumentation-client.ts, and the sentry.server/edge/shared config files are all present and Sentry is used across routes. Closing as completed.

Issue #474, opened 2026-03-08
[RESILIENCE] Critical email retry with dead-letter queue


teetangh:
## Problem

Critical emails (payment success/failure, password reset, booking confirmations) are fire-and-forget. If Resend is down or rate-limited at the moment of sending, the email is silently lost with no retry or record.

## Proposed Solution

1. **Retry critical emails** — 3 attempts with exponential backoff (1s, 4s, 16s)
2. **Dead-letter table** — `FailedEmail` model to persist failed emails for manual retry:
   ```
   model FailedEmail {
     id          String   @id @default(cuid())
     to          String
     subject     String
     templateId  String
     payload     Json
     attempts    Int      @default(0)
     lastError   String?
     retriedAt   DateTime?
     createdAt   DateTime @default(now())
   }
   ```
3. **Admin dashboard** — View and retry failed emails
4. **Alert threshold** — Novu notification to admin when email failure rate > 5% in 1 hour

## Acceptance Criteria

- [ ] Critical emails retry 3x before giving up
- [ ] Failed emails persisted in `FailedEmail` table
- [ ] Admin can view and retry failed emails
- [ ] Alert fires when failure rate exceeds threshold
- [ ] Non-critical emails (marketing, updates) remain fire-and-forget

## Key Files

- `lib/email.ts` — Current fire-and-forget email sending
- `lib/novu/service.ts` — Notification pattern reference

teetangh:
Verified implemented. model FailedEmail (attempts, nextRetryAt, EmailDeliveryStatus) plus recordFailedEmail in lib/email.ts, the worker jobs/email/retry-failed-emails.ts, its cron, and admin requeue of DEAD_LETTER via app/api/admin/failed-emails/route.ts. Closing as completed. Optional future enhancement: a failure-rate alert threshold.

Issue #473, opened 2026-03-08
[RESILIENCE] Stream.io circuit breaker and graceful degradation


teetangh:
## Problem

When Stream.io experiences an outage, calls to the Stream API throw unhandled errors that propagate to users as white screens or cryptic error messages. There is no fallback or circuit breaker.

## Proposed Solution

1. **Circuit breaker** wrapping the Stream client — Same pattern as Redis circuit breaker in `lib/redis.ts`
   - States: CLOSED (normal), OPEN (failing, skip calls), HALF-OPEN (testing recovery)
   - Track failure count over sliding window
   - Open circuit after N consecutive failures
2. **Graceful degradation** — On Stream outage:
   - Show "Video service temporarily unavailable" instead of error
   - Disable "Join" buttons with explanation
   - Cache last-known-good Stream tokens briefly during outage
3. **Health check** — Periodic Stream API ping to detect recovery

## Acceptance Criteria

- [ ] Circuit breaker wraps all Stream API calls
- [ ] Open circuit shows user-friendly message instead of error
- [ ] Circuit auto-recovers when Stream comes back
- [ ] Token caching during brief outages
- [ ] Health endpoint includes Stream status

## Key Files

- `lib/stream-client.ts` — Stream singleton to wrap
- `lib/redis.ts` — Circuit breaker pattern reference
- `app/api/health/route.ts` — Add Stream status check

teetangh:
## Partial Progress — PR #532 (merged)

PR #532 addressed several resilience items that overlap with this issue, though the full circuit breaker pattern is not yet implemented:

**What was done:**
- ✅ Exponential backoff retry with "Retrying connection (attempt N/5)" UI (StreamProvider)
- ✅ StreamErrorBoundary with retry mechanism (pre-existing, verified working)
- ✅ Webhook retry support — failed events can now be re-processed instead of silently dropped
- ✅ Recording stop rollback — DB state reverts on Stream API failure
- ✅ Transfer failure recovery — recordings revert to READY instead of permanently FAILED
- ✅ Token caching with per-type expiry prevents unnecessary token refreshes

**Still needed for full close:**
- [ ] Circuit breaker wrapping all Stream API calls (CLOSED → OPEN → HALF-OPEN pattern)
- [ ] "Video service temporarily unavailable" graceful degradation UI
- [ ] Health endpoint Stream status check
- [ ] Token caching during brief outages

teetangh:
[...]

PR #877, opened 2026-06-16
fix(documents,stream,email,collaborators): launch-blocker productionization (batch 2)

teetangh:
## Subsystem productionization — launch-blockers (batch 2 of 2)

Second of two PRs. **Stacked on #876** (base branch = `fix/subsystems-money-access-correctness`) so the diff here is only this batch; GitHub auto-retargets to `dev` once #876 merges. Carries the remaining launch-blockers for Documents, Stream/meetings resilience, the email dead-letter queue, and the robust collaborator double-booking guard. Scope was confirmed item-by-item after a necessity-research pass (some originally-listed items were dropped as already-handled — see below).

### Documents (#694)
- **DOC-1** — Document download and listing now reject once the parent booking is terminal (cancelled / rejected / expired), closing a post-refund visibility gap. Implemented via a typed `isBookingTerminal()` helper (`lib/appointments/terminal-status.ts`) over the real child status enums — `Appointment` has no status of its own.
- **DOC-2** — Document uploads are now rate-limited (`documentUploadLimiter`), reusing the existing `applyRateLimit`/`makeLimiter` pattern, on both the consultee and consultant upload routes.
- **DOC-3** — The storage-reconcile script now persists missing-file state on the new `AppointmentDocument.isStorageMissing` / `missingDetectedAt` columns (and clears it on recovery) instead of only logging; the previously hard-coded `missingFilesMarked: 0` is now real.

### Stream / meetings (#689, resilience)
- **#473** — Hot-path Stream calls are wrapped in the shared `withCircuitBreaker` (`withStreamCircuitBreaker`) so a Stream outage fast-fails in sub-ms instead of every dashboard load eating the 30s client timeout and cascading. Closed-breaker behaviour is unchanged; open-breaker degrades (empty channel list) or throws a typed `StreamUnavailableError`.
- **STR-4** — New `MeetingAttendance` model plus `call.session_participant_joined`/`left` webhook handlers capture per-attendee join/leave (`firstJoinedAt`/`lastLeftAt`/`joinCount`). This is the data that **unblocks no-show (#471) and overrun (#472) detection**, which previously had nothing to read.
- **STR-2/3** — Recording-transfer failures are now tracked (`transferAttempts`/`lastTransferError`) and page engineering once via `recordSystemError` after repeated failures (deduped by `transferFailureAlertedAt`); the previously-TODO `STREAM_ONLY` expiry-warning email now actually sends through Novu.

### Email dead-letter queue (#474)
- A failed Resend transactional send now persists the **already-rendered** message to a new `FailedEmail` table (typed columns — no Json escape hatch) instead of being silently dropped. A retry worker (`jobs/email/retry-failed-emails.ts`, new cron) re-sends verbatim on a 1m/5m/30m/2h/8h backoff and dead-letters after exhaustion, with an operator replay route. Mirrors the proven `OutboundWebhookDelivery` worker.

### Collaborators (#784, AE-2 — robust layer)
- The plan owner's `consultantProfileId` is now denormalized onto webinar/class **group-event** slots (previously NULL there), so the `slot_no_confirmed_overlap` exclusion constraint finally guards the host against double-booking across **all** event types — a clash surfaces as Postgres `23P01` → HTTP `409`. The co-host availability guard from #876 is retained.

### Observability (#475)
- No Sentry here. Documented that the existing Better Stack sink (`recordSystemError` → `recordSystemEvent`, gated by `ENABLE_BETTERSTACK_TELEMETRY`) already pages on the critical money/webhook/cron paths; **full Sentry is deferred to a separate PR** per the agreed scope.

### Dropped after necessity research (not regressions)
- **#248 chat-sync "storm"** — the "50–100 calls every load" claim is stale; sync is already gated to once-per-session with TTL caches. Only a power-user first-sync fan-out remains (a perf nicety), so it was left out.
- The recording **expiry-vs-cron race** was found not real (5-day transfer lead vs a 6-hour cron); only the genuinely-missing pieces (retry tracking, paging, expiry email) were done.

### Verification
- `tsc --noEmit`: 0 errors
- jest: 27 suites / 261 tests pass (incl. new suites for DOC-2 limiter, email retry/backoff, attendance handlers, transfer-failure paging; existing stream/payments/referrals/collaborators suites green)
- No `as any` / `unknown` / `typeof`-cast patterns in the diff (per review feedback); the one stream-chat-v9 SDK custom-data cast is pre-existing and unavoidable.

### Deployment notes (external config, not code)
- The Stream dashboard webhook subscription must include `call.session_participant_joined` / `call.session_participant_left` for STR-4 attendance to flow.
- Schema additions (`FailedEmail`, `MeetingAttendance`, the new columns) apply on the next `prisma db push` — not pushed here, per the shared-dev-DB convention.

Part of #689, #692, #694, #471, #472, #473, #474, #475, #784.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -43,7 +43,9 @@ NEXT_PUBLIC_TEST_USERID=""
 # Redis connection URL
 REDIS_URL=""
 
-# Resend API key for transactional email service
+# Resend API key for transactional email service. #474 — also drives the
+# email retry worker (jobs/email/retry-failed-emails.ts), which re-sends
+# rows the senders dead-lettered to FailedEmail on a transient Resend failure.
 RESEND_API_KEY=""
 
 # Razorpay API credentials for payment processing
@@ -132,12 +134,13 @@ SUPPLIER_STATE_CODE="KA"
 # consolidated billing. See app/api/cleanup/consolidated-invoice-rollup/.
 ENABLE_CONSOLIDATED_INVOICE="false"
 
-# #776 §K — Better Stack Telemetry (logs) sink for operational events. When
-# "true" AND the source token + ingest URL are set, recordSystemEvent/
+# #776 §K / #475 — Better Stack Telemetry (logs) sink for operational events.
+# When "true" AND the source token + ingest URL are set, recordSystemEvent/
 # recordSystemError ALSO ship to Better Stack so a failed reconcile / stuck
 # payout / webhook-queue backlog / HMAC failure pages someone. The DB
 # SystemEvent row is always the source of truth; this is a best-effort side
-# channel. Create a Telemetry source in Better Stack to get the token + host.
+# channel. Set ENABLE_BETTERSTACK_TELEMETRY="true" + both vars below to turn
+# it on. Create a Telemetry source in Better Stack to get the token + host.
 ENABLE_BETTERSTACK_TELEMETRY="false"
 BETTERSTACK_SOURCE_TOKEN=""
 BETTERSTACK_INGEST_URL=""
--- a/.github/workflows/retry-failed-emails.yml
+++ b/.github/workflows/retry-failed-emails.yml
@@ -1,0 +1,51 @@
+name: Retry Failed Emails
+
+# #474 — transactional-email dead-letter retry. lib/email.ts persists a rendered
+# FailedEmail row when a Resend send fails; this worker re-sends due rows, walks
+# the 1m/5m/30m/2h/8h backoff, and dead-letters once exhausted (operator-replayable
+# via the admin route). Every 15 minutes is enough granularity for the backoff
+# slots without churning a low-volume failure table.
+
+on:
+  schedule:
+    - cron: "*/15 * * * *"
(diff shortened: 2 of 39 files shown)


teetangh:
Addressed the Gemini review findings in 06962468:

**HIGH — Resend resolves with `{ error }` rather than throwing on API-level failures (rate limit, invalid key, validation).** This was a real correctness bug in the DLQ: the catch-only logic would have missed those, and the retry worker would have marked such a send as `SENT`. Fixed at the root — all seven `lib/email.ts` senders now `throw` when `data.error` is non-null so the existing catch dead-letters the rendered message, and `runEmailRetryTick` treats a returned `error` as a failure (RETRY → DEAD_LETTER) instead of a success. Added a worker test that pins the returned-error path.

**MEDIUM — reconcile N+1.** `persistMissingState` now batches the missing/recovered flag writes into two `updateMany` calls instead of one `update` per row.

tsc clean; email worker + document-limiter suites pass (10 email tests incl. the new returned-error case).

PR #926, opened 2026-06-26
fix(csp): allow Novu WebSocket connections

ehtashammd07:
## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded allowed network connections so the app can communicate with additional Novu service endpoints.
  * Improved support for secure WebSocket and HTTPS connections to related services, helping prevent connection-related issues in the browser.


next.config.mjs
```diff
@@ -35,7 +35,8 @@ const withBundleAnalyzer =
 const CSP_DIRECTIVES = [
   "default-src 'self'",
   "script-src 'self' 'unsafe-inline' 'unsafe-eval' https://checkout.razorpay.com https://*.sentry.io https://*.getstream.io https://*.supabase.co",
-  "connect-src 'self' https://*.getstream.io wss://*.getstream.io https://*.supabase.co https://*.upstash.io https://api.razorpay.com https://*.sentry.io https://api.resend.com",
+  // "connect-src 'self' https://*.getstream.io wss://*.getstream.io https://*.supabase.co https://*.upstash.io https://api.razorpay.com https://*.sentry.io https://api.resend.com",
```
teetangh: should-fix (tiny): this is a commented-out copy of the old line — let's delete it rather than leave it in. Git already preserves the previous version in history, and a stray commented directive tends to confuse the next person editing this allow-list ("is this meant to be re-enabled?"). Dropping it keeps the +2/-1 down to a clean +1/-1.

next.config.mjs
```diff
@@ -35,7 +35,8 @@ const withBundleAnalyzer =
...
+  "connect-src 'self' https://*.getstream.io wss://*.getstream.io https://*.supabase.co https://*.upstash.io https://api.razorpay.com https://*.sentry.io https://api.resend.com https://*.novu.co wss://*.novu.co",
```
teetangh, line 39: nit (optional): since `NovuProvider.tsx` doesn't set a custom `backendUrl`/`socketUrl`, we only actually hit `api.novu.co` + `ws.novu.co` today, so you *could* pin the exact hosts instead of `*.novu.co` for a slightly tighter policy. That said, the wildcard matches the style of the other entries here and keeps us covered if we ever switch to the EU region (`eu.api.novu.co` / `eu.ws.novu.co`), so I'd happily keep it as-is — just flagging the trade-off. Could you double-check in the browser Network tab which `*.novu.co` host the inbox actually opens the WS to, so we're 100% sure the wildcard covers it?

teetangh (commented):
Nice work on this — you picked the right directive and the right scheme, which is the part people most often get wrong with WebSocket CSP. 🎯

What's good here:
- **Correct directive + scheme.** Novu's realtime feed runs over a WebSocket, so it belongs in `connect-src` with a `wss://` origin — exactly what you did. It would NOT work under `script-src`/`frame-src`, so good instinct.
- **Completeness.** You added *both* `https://*.novu.co` (Novu's REST API) and `wss://*.novu.co` (the socket). The Novu inbox SDK talks HTTP first and then upgrades to WS, so it genuinely needs both entries — easy to forget the HTTPS one. Well spotted.
- **Sensible scoping.** `*.novu.co` is a host-scoped wildcard, not an over-broad `wss://*`. Since `providers/NovuProvider.tsx` uses the default Novu Cloud endpoints (no custom `backendUrl`/`socketUrl`), this covers both the US/global hosts (`api.novu.co`, `ws.novu.co`) and the EU hosts (`eu.api.novu.co`, `eu.ws.novu.co`) — so we're future-proof if we ever move to the EU region. It's also consistent with the existing getstream/supabase/sentry/upstash wildcards in this list.

A couple of small things below — nothing blocking, mostly cleanup. One heads-up for testing: this header is served as `Content-Security-Policy-Report-Only` unless `ENABLE_CSP_ENFORCE=true` (see `CSP_HEADER_KEY` just below this block), so on the current deploy the *old* policy wouldn't actually have blocked the Novu socket — it'd only have logged a violation to `/api/csp-report`. Worth confirming the inbox actually connects (Network tab → WS) rather than relying on the absence of console errors. 🙂

Thanks for tightening this up!

PR #944, opened 2026-06-27
chore: sync prod → dev (re-align history after #940/#943 promotions)

teetangh:
Routine **prod → dev back-merge**, matching the existing pattern (e.g. #928).

`prod` is ahead of `dev` only by the two dev→prod promotion **merge commits** (#940 instant-navigation, #943 build-OOM fix) — the branches are **content-identical** (zero file diff). This PR re-absorbs those merge commits into `dev` so the history stays aligned and future dev→prod release diffs remain clean.

No code change — history alignment only. Merge with a **merge commit** (not squash); do **not** delete the `prod` branch.


teetangh:
Closing — the prod→dev back-sync can't merge while dev has the un-promoted #939 (dev protection enforces require-up-to-date + enforce_admins, so no bypass is possible). dev already contains all of prod's code; the divergence is cosmetic history structure only and will re-align on a future promotion when dev is content-equal to prod. No action needed.

Issue #937, opened 2026-06-27
infra: evaluate Prisma Accelerate (ap-south-1 pool) as a durable cross-region fix — relates to #932


teetangh:
## Summary

This issue evaluates [Prisma Accelerate](https://www.prisma.io/docs/accelerate) as a durable remedy for the cross-region database latency documented in #932, where our Netlify Functions execute on AWS Lambda in `us-east-2` (Ohio) while our Supabase Postgres and its Supavisor pooler live in `ap-south-1` (Mumbai). Accelerate replaces the direct cross-region Postgres connection with a managed connection pool that we place in `ap-south-1` (Mumbai), right next to the database, so the Lambda issues an HTTP request to Accelerate instead of paying for a cold cross-region Postgres handshake on the hot path. The attraction of this path is that it removes the cold-connect timeout as a class without migrating the database and without leaving Netlify, at the cost of a code change in `lib/prisma.ts`, a usage-based bill, and a managed proxy on the query path. This is a sibling, durable-option follow-up to #932; it does not replace the resilience and caching work already shipped under #929, #931, #934 and #935, which contain the blast radius but do not remove the latency.

## Why this is needed

The mitigations already shipped under #932 make the failure mode graceful but leave the underlying physics in place. Per-route fail-open resilience (#929, #931, #934) degrades a timed-out request to fallback data instead of a crash, and the data-layer `unstable_cache` of public reads (#935, "#932 Phase 2") keeps the hottest public surfaces off the pooler. Neither removes the cold cross-region Postgres connect that causes `FAMILIARISE_WEB-A` and `FAMILIARISE_WEB-9` in the first place; they only bound its consequences. A durable fix has to put the compute and the database in (or effectively next to) the same region, or remove the cold-connect step entirely.

[...]

Issue #932, opened 2026-06-26
infra: cross-region latency (Netlify us-east-2 ↔ Supabase ap-south-1) causes Postgres pooler connection timeouts


teetangh:
## Summary

Two production errors reported by Sentry (org `practitionist`, project `familiarise_web`) are Prisma/Postgres pooler connection timeouts that originate in our network topology rather than in application logic. Our Netlify Functions execute on AWS Lambda in `us-east-2` (Ohio), while our Supabase Postgres database and its Supavisor pooler live in `ap-south-1` (Mumbai). The resulting cross-region round-trip of roughly 200 ms or more means that an occasional cold TLS-and-authentication handshake to the pooler exceeds our deliberately tight 3-second runtime connect budget, which surfaces to users as a failed request. This issue documents the evidence, confirms the root cause, records what we have already done to contain the blast radius, evaluates every viable remedy against our specific stack, and proposes a concrete recommended fix.

## Evidence

The two production Sentry issues are both pooler connection timeouts whose stack frames bottom out in the node-postgres connection path (`PrismaPgAdapter.performIO` calling into `pg-pool` and timing out on connect). The platform context for both is Node.js v22 running on AWS Lambda (these are Netlify Functions).

- **`FAMILIARISE_WEB-A`** — `Error: timeout exceeded when trying to connect`. This is thrown from the Next.js Server Component for `/explore/experts`. When it was unhandled it propagated to a full page crash. It has 7 events and was first seen roughly six hours before this issue was filed.
- **`FAMILIARISE_WEB-9`** — `Error: Connection terminated due to connection timeout`, wrapping a nested `Connection terminated unexpectedly`. This is thrown from `GET /api/announcements`. It has 10 events.

The decisive piece of evidence is the region mismatch captured in the Sentry event context. The function context reports `cloud.region: us-east-2` and `cloud.platform: AWS_Lambda_nodejs22.x`, confirming the Netlify Function runs in Ohio. The application's `DATABASE_URL` host is `aws-0-ap-south-1.pooler.supabase.com:6543`, confirming the Supavisor transaction pooler is in Mumbai. The two endpoints are read-only public surfaces (`app/explore/experts/page.tsx` reads through `lib/data/explore-experts`, and `app/api/announcements/route.ts` is a public `GET` running `prisma.announcement.findMany`), and neither currently declares any caching, so each request opens a fresh cross-region pooled connection on the hot path.

## Root cause

[...]

Issue #900, opened 2026-06-19
Sentry observability: current landscape + follow-ups


teetangh:
## Overview

Sentry was wired into the application via the official `@sentry/nextjs` wizard on the branch `feat/sentry-observability`. This issue records the **entire current Sentry landscape** and tracks the **remaining follow-up work**. The initial wiring lands in the accompanying PR; everything in the checklist below is intentionally deferred so the foundation can ship first.

## Current landscape (what is in place)

The SDK is `@sentry/nextjs` (v10) reporting to the SaaS project **`practitionist/familiarise_web`** in the US region. Initialization is split per Next.js runtime and registered through `instrumentation.ts`.

| Area | State |
| --- | --- |
| Error monitoring | **Enabled** across server, edge, and browser runtimes |
| Performance tracing | **Enabled** (`tracesSampleRate: 1` — currently 100% everywhere) |
| Sentry Logs | **Enabled** (`enableLogs: true`) |
| Session Replay | **Disabled** on purpose — keeps the client bundle light and shrinks the DPDP surface on PII-heavy consult screens |
| Ad-blocker tunnel | **Disabled** on purpose — avoids per-event Netlify function cost; left commented in `next.config.mjs` for later |
| Source maps | Uploaded at build by `withSentryConfig`, gated on `SENTRY_AUTH_TOKEN` + org/project (no-ops without them) |
| Server-error capture | `onRequestError` in `instrumentation.ts` |
| Render-crash boundary | `app/global-error.tsx` |
| Verify scaffolding | `/sentry-example-page` + `/api/sentry-example-api` (temporary) |

The relevant files are `instrumentation.ts`, `instrumentation-client.ts`, `sentry.server.config.ts`, `sentry.edge.config.ts`, `app/global-error.tsx`, and the `withSentryConfig` wrapper in `next.config.mjs`. The Content-Security-Policy in `next.config.mjs` already allow-lists `*.sentry.io`, so no CSP change was required.

Environment variables are present in both local `.env` files (gitignored): `NEXT_PUBLIC_SENTRY_DSN`, `NEXT_PUBLIC_SENTRY_ENVIRONMENT`, `SENTRY_ORG`, and `SENTRY_PROJECT`. The secret `SENTRY_AUTH_TOKEN` lives only in `.env.sentry-build-plugin` (also gitignored). The added client bundle weight is roughly 35–45 KB gzipped; declining Session Replay avoided an additional ~50 KB or more.

[...]

Issue #734, opened 2026-05-01
Frontend performance: Prisma over-fetch, zero Next.js caching, raw fetch() everywhere, oversized client bundles


teetangh:
## Symptom

Across landing, explore, checkout, and especially the dashboards (consultee, consultant, organization), navigation between pages is **noticeably slow** and dashboard data fetching adds another layer of perceived lag. This issue collects the systemic causes (not individual slow queries) and lays out a phased fix plan.

## Quick audit numbers (May 2026)

Run against `feature/enterprise` branch:

| Signal | Count | What it tells us |
|--------|-------|------------------|
| API route handlers | **328** | Heavy REST surface |
| Client components (`"use client"`) | **279** | Most UI is rendered client-side |
| `useQuery` / `useMutation` sites | **77** | TanStack Query is the data layer |
| Raw `fetch()` calls in `app/` + `components/` | **133** | A third of fetches bypass TanStack — no dedup, no cache |
| Server Actions (`"use server"`) | **12** | Almost unused |
| `unstable_cache` + `revalidateTag` + `revalidatePath` | **0** | Zero Next.js native data cache |
| React `cache()` wrappers | **5** | Minimal request-scoped memoization |
| Prisma `include:` occurrences | **1357** | Default to over-fetch |
| Prisma `select:` occurrences | **875** | Used about 60% as often as `include` |
| `Suspense` usages | **9** | Almost no streaming UI |
| `loading.tsx` route segments | **12** | Out of ~100 routes |
| Pages with `force-dynamic` / `revalidate` | **2** | No ISR / static caching |
| Client components > 500 LoC | **10+** | Big bundles ship to every browser |

### 1. Default-`include` over-fetching in Prisma

The codebase reaches for `include:` 1.5× more often than `select:`. Bare `include` pulls **every column** of the joined model, including big text/JSON columns and (until last week) Decimal fields that triggered Server→Client serialization errors.

The hottest routes have monster includes:

```
84 includes  app/api/slots/appointments/[appointmentId]/route.ts
54 includes  app/api/events/consultations/[consultationId]/route.ts
50 includes  app/api/events/subscriptions/[subscriptionId]/route.ts
37 includes  app/api/dashboard/consultant/[consultantId]/requests/route.ts
34 includes  app/api/slots/appointments/route.ts
```

[...]

PR #945, opened 2026-06-30
fix: Sentry stabilization + config/storage/auth debloat

teetangh:
## Context

This PR began as a request to audit the Next.js / Netlify / Supabase setup for complexity and bloat, and to fix every unresolved Sentry issue in one pass. The audit's headline finding is that the core setup is largely well-engineered and intentional rather than bloated, and that the stack is actually three separate concerns — BetterAuth for authentication, Supabase Storage for files, and Prisma over Supabase Postgres for data — not a Supabase-Auth application. Sentry had four unresolved issues, not the large backlog we expected: two are cross-region database-pooler timeouts whose root cause is infrastructure (tracked in #932 / #937), and two are pure third-party noise. This PR fixes and closes all four, performs the genuinely safe debloat the audit surfaced, collapses the one real bloat hotspot (`lib/supabase.ts`), and makes a small, safe reduction to the authentication hot path.

## Workstream A — Sentry: fix and close all four issues

The two cross-region timeouts (`FAMILIARISE_WEB-9` on `GET /api/user/consultants` and `FAMILIARISE_WEB-A` on `/explore/experts/[consultantId]`) now fail open. The API route branches its catch on the existing `isTransientDbError` helper and degrades to an empty, cache-headed `200` with a breadcrumb instead of capturing an exception and returning a 500, mirroring the pattern already shipped for the announcements banner (#931). The expert detail Server Component catches the same transient class inside render so it can never escape to `onRequestError`, rendering a retriable degraded state; a new segment `error.tsx` (the app previously had none) covers genuine defects. The two noise issues are filtered at the single shared `Sentry.init`: `ignoreErrors` and `denyUrls` drop the injected wallet-extension errors (`inpage.js`, `FAMILIARISE_WEB-F`) and the React Server Components stream-abort (`"Connection closed."`, `FAMILIARISE_WEB-E`). The Prisma timeout strings are deliberately not filtered, so a new unprotected route surfacing them remains detectable.

**Sentry dashboard actions.** `FAMILIARISE_WEB-9` and `-A` have already been muted (until-escalating) with comments linking #932 / #937, because their root cause is infrastructure and persists; they will be resolved for real once the durable fix lands. `FAMILIARISE_WEB-E` and `-F` will be resolved once this PR deploys and the filters take effect, since resolving them earlier would auto-regress as events keep arriving against the un-filtered build.

## Workstream B — Safe config debloat

The highest-value item here closes a latent production risk: the Content-Security-Policy did not allow-list Stripe even though Stripe is live (Connect payouts, checkout, and webhooks), so enforcing the CSP (`ENABLE_CSP_ENFORCE=true`) would have broken payouts and checkout. Stripe's domains are now present in `script-src` / `connect-src` / `frame-src`. The remainder is targeted cleanup: an orphan `@types/redlock` dependency, the `build:prod` alias and the deprecated `next lint` script, the Vercel-only `automaticVercelMonitors` flag, the no-op Netlify context blocks, and a dead Tailwind `./pages/**` glob are all removed; `.env.sample` now documents two used-but-missing variables (`SUPABASE_SERVICE_ROLE_KEY`, `NEXT_PUBLIC_SENTRY_ENVIRONMENT`) and drops a dead one; a `.nvmrc` pins local Node to 22; the Prisma client no longer falls back to `DIRECT_URL` at runtime (which could have silently routed traffic onto the un-pooled direct connection); and a broken adapter-less test client was deleted.

## Workstream C — Collapse `lib/supabase.ts`

`lib/supabase.ts` was the one genuine bloat hotspot, at 1844 lines of roughly thirteen near-identical upload / delete / list helper families. It is now 1113 lines (−40%): the families call a small generic core (`uploadAsset` / `deleteAsset` / `deleteAssetFolder` / `listAssets` / `getPublicAssetUrl` / `getSignedAssetUrl`), every public function keeps its exact name, signature, and behaviour, and the nineteen consumer files are unchanged. Verified-dead exports (the cover-image family, two list helpers, the deprecated `ensureFolderExists`, and four unused constants) were deleted. A `server-only` guard was added (safe, because all client importers use `import type`), and `ensureBucketExists` now memoizes per process instead of issuing a `.list()` round-trip on every upload. The behavioural invariants that matter — signed-URL TTLs, the signing-client divergence (appointment documents use the admin client with an anon fallback, while plan materials and consultant documents sign anon-only), per-family MIME and size limits, the upsert and folder-replace flags, the rich-versus-slim result shapes, the org SVG filename path, and every storage-tagged Sentry capture — are preserved exactly.

## Workstream D — Authentication hot-path reduction

`customSession` runs on every protected API request and onboarding-guard render (these paths intentionally disable the cookie cache to keep roles and onboarding state fresh). It previously issued two separate up-front round-trips; the bare-member lookup is now folded into the session `user` query via a filtered relation include, removing one cross-region round-trip per resolution with no change to the data or its freshness. The more aggressive `sessionGeneration`-keyed membership cache that the code scaffolds was deliberately not implemented, because the session payload carries `walletBalance` and `fundingSource` (money-display data) that the generation marker does not invalidate; caching it would serve stale money, which violates our "never cache money truth" rule. That deeper optimisation is left to #734, where the money fields would first move off the hot payload.

## Verification

`tsc --noEmit` is clean (zero errors, run cold with a raised heap to avoid the known incremental trap). The storage and auth-membership Jest suites pass (7 / 7). IDE diagnostics are clean on every hand-edited file, and the storage refactor's critical paths were reviewed line-by-line against the invariant checklist. A production-mode `next build` and a live dev-server smoke (happy paths plus a forced-transient degradation check on the two fail-open routes) are being run before merge; live authenticated per-family uploads were not smoke-tested and rest on the static verification above.

## Out of scope (tracked elsewhere)

The durable cross-region fix (Prisma Accelerate, a Netlify region change, or a Supabase migration) remains tracked in #932 / #937. Setting `PG_POOL_MAX=1` in the Netlify runtime environment is an operational follow-up that cannot live in `netlify.toml`.

Part of #900, #932, #733, #734.

## Summary by CodeRabbit

* **New Features**
  * Added a friendly “profile couldn’t be loaded” fallback for expert pages, with navigation back and a retry option.
  * Updated the environment sample to support server-side Supabase credentials for admin/signed downloads.

* **Bug Fixes**
  * Improved resilience to temporary backend connection issues for expert pages and consultant listings (graceful degraded UI/empty results instead of hard failures).
  * Reduced non-actionable error noise from known third-party/browser-extension sources.

* **Chores**
  * Updated deployment/runtime configuration (Node version), monitoring setup, and tightened/expanded CSP for external vendors.


teetangh:
### Verification — all green ✅

- **`tsc --noEmit`**: clean, 0 errors (run cold with an 8 GB heap to avoid the incremental trap).
- **Jest**: storage + auth-membership suites — **7/7 passed**.
- **Production `next build`** (NETLIFY mode): **exit 0**, every route compiles, no `server-only` client-bundle errors, `/explore/experts/[consultantId]` correctly `ƒ (Dynamic)`.
- **Dev-server smoke** (transient forced via an unreachable `DATABASE_URL` → real 3 s connect-timeout):
  - `GET /api/user/consultants` → **`200`** + `{"data":[],"meta":{"total":0,"page":1,"limit":10,"totalPages":0}}` + breadcrumb `[fail-open] consultants list read … degrading` — degrades, does **not** capture or 500.
  - `/explore/experts/[id]` → **`200`** degraded "taking a moment to load" + "Try again", breadcrumb `[fail-open] consultant detail page … degrading` — the transient never escapes render to `onRequestError`.
  - Happy path: `GET /api/user/consultants` → `200` + data.

Live authenticated per-family uploads were **not** smoke-tested (they need a logged-in session); they rest on the clean `tsc`, the storage Jest suite, and a line-by-line review of the signing-client / TTL / MIME / size invariants.

Sentry: `FAMILIARISE_WEB-9` and `-A` are already **muted** (until-escalating) with comments linking #932 / #937; `-E` and `-F` will be **resolved** once this deploys and the filters take effect.

teetangh:
## Update — review fixes + the "No experts found" diagnosis

Pushed `9553b902` addressing this PR's review, plus a regression the review surfaced on the deploy preview.

**Bot comments triaged (9):**
- ✅ **Fixed:** the fail-open `Cache-Control` (→ `no-store`; see below), the `subsystem` tag (→ `consultants`), `unstable_rethrow()` in the expert-detail catch so `notFound()`/`redirect` can't be swallowed, and best-effort orphan-object cleanup on signed-URL failure in `lib/supabase.ts`.
- ⏭️ **Skipped (BS / pre-existing, justified):** the two Gemini `storagePath` "double-bucket" claims (byte-identical to the working production code — `pathWithinBucket` doesn't exist), the "use service-role" suggestion (the refactor preserves the original anon-client behavior), and two pre-existing list-pagination / bucket-MIME-allowlist notes.

**"No experts found" on the deploy preview — root cause + fix:** the fail-open branch returned the empty `[]` with the *success* path's `s-maxage=60, stale-while-revalidate=300` header, so a single cross-region transient timeout got the empty result **CDN-cached** and served for minutes (a cache-busted origin request returns all 83 experts; env + DB are byte-identical prod↔preview per a per-context check). Fixed with `no-store` on the fail-open branch, plus a default-page `unstable_cache` for perf.

**Scope split:** the consultant-PII exposure this review surfaced turned out to be **systemic** (13 consultant-serving surfaces leaking PAN / bank / TDS data). To keep this PR focused, that remediation moved to a dedicated security PR → **#946**. This PR keeps only the list-route fix (via the `IConsultantCardData` allowlist, the stronger pattern).

Issue #1021, opened 2026-07-18
Enterprise org-awareness gaps: org-scoped support, feedback, and quality/review signal


teetangh:
From the B2C/B2B domain-boundary audit (PR #1019 follow-up work). The audit found **no mismatches** — trials, referrals, and discount codes are correctly B2C, and the operational tags (Appointment/Payment/Waitlist/Recording/MeetingSession/Plans) are correctly cross-cutting. The open items are all **additive**: B2C-only concepts an enterprise customer would reasonably expect to be org-aware. To be built as **separate follow-up PRs** after the cross-cutting identity + Org Appointments PR merges.

### 1. Org-scoped support tickets (highest value)
`SupportTicket` (prisma/schema.prisma ~:167-201) links to consultation/subscription/payment but carries **no `organizationId`**, so an org admin can't see/triage a sponsored member's ticket about an org-funded booking, and staff can't route it to the org.
- Add `organizationId String?` to SupportTicket; populate from the linked payment/appointment at create.
- Add an org-scoped list route mirroring `lib/api/scope/list-*.ts`.
- Add a dashboard page under `app/dashboard/organization/[orgId]/`. Self-contained.

### 2. Org-scoped feedback
`Feedback` (~:150-165) is `userId`-only — no rollup for the enterprise buyer. Same fix shape as #1, lower priority.

### 3. Org-level quality / review signal
`ConsultantReview` is consultant↔consultee marketplace trust only, and **webinar/class sessions have no review path at all** (`lib/reviews.ts:34-80` covers only consultation/subscription/trial). An org has no satisfaction/NPS signal for the sessions it sponsors or hosts.
- Do NOT overload the public `ConsultantReview`; add a new org-aware feedback/CSAT model + read surfaces.
- Extending eligibility to webinar/class touches `hasCompletedBookingWith`.

(A 4th, lower-priority gap — org-scoped notifications — has no existing entity to extend and would be fully net-new; deferred.)

Part of #746 #867

teetangh:
Vocabulary note (2026-09-14): since #1638 the scheduling models this issue names are renamed — `MeetingSession` → `Meeting`. The body is left as written; read it through the retired-name table at the end of `docs/enterprise/00-foundations/07-scheduling-glossary.md` (`scripts/ci/check-terminology.ts` refuses the old names in code).

PR #1032, opened 2026-07-27
fix(scope): stop default scopes silently hiding org-funded rows

teetangh:
Fourth instance of one pattern: a surface resolves a default scope, that default
excludes org-funded rows, and the surface renders a subset while looking
complete. Nothing errors and no row is marked missing. `consultee/home` and
`consultee/payments` were fixed when found; these two were not.

## 1. Slot-allocation requests never reached the expert

`RequestSlotAllocationTab` is the **only** surface in the product that allocates
slots — there is no requests or allocation page anywhere under
`app/dashboard/organization/`. It fetched `/api/bookings/consultations` and
`/api/bookings/subscriptions` with no `orgScope`, and both routes apply this
when the param is absent:

```ts
const defaultPersonal = !rawOrgScope && !isPrivileged(role);
if (explicitPersonal || defaultPersonal) { /* excludes org-funded rows */ }
```

So an org-sponsored subscription was paid for and then **never scheduled**. The
request to allocate its slots existed and appeared on no surface that could act
on it. Money moved; the sessions did not happen.

Four call sites now send `orgScope=all` — the requests tab, its home-page mini
variant, the shared consultant `fetchHelpers`, and `useEvents`. All four read
the caller's OWN rows and the routes are self-scoped by profile id, so `all`
means "all of mine" (`allowAllForOwner`), not a widening.

## 2. The chat inbox was pinned to one org

`ChatSidebar` called `useOrgScope()` bare. That defaults to `first-org`, so the
channel query became `organization_id: { $eq: firstOrgId }` — a member's B2C
conversations disappeared behind their first org, and a member of two orgs never
saw the second one's at all. With no way to change it, since ADR 19 removed the
personal-dashboard context filter and nothing renders a scope picker.

The channel topology needed nothing: `messaging` DMs for consultations and
subscriptions, `team` channels for webinars and classes, all tagged with
`organization_id` by the shared creator, and the sidebar already queried both
types. Only the filter was wrong.

**Chat stays out of the organization dashboard and the org still cannot read any
of it — ADR 20 is untouched.** This is about a member reaching their own
conversations.

## Why the call sites, not the defaults

The right default genuinely differs by surface: a spend report wants personal,
an inbox wants everything. Flipping `resolveOrgScope` would relocate the bug
into the surfaces #1029 spent its length narrowing.

`__tests__/security/org-scope-defaults.test.ts` pins the class instead — every
`useOrgScope` caller must state `defaultForOrgMember`, and every caller of the
two org-excluding endpoints must send `orgScope`, checked per line so one
correct and one forgotten call in the same file still fails.

## Verification

`tsc` clean · 1807 tests / 157 suites · lint clean on the diff. The new test was
confirmed failing 4/11 against the pre-fix code.

Audit scope, per the agreed approach: 4 `useOrgScope` call sites (3 already
correct), 2 routes carrying the excluding default, 4 callers of those routes.
All now explicit.

No schema change, so no DB push.

## Not in this PR

The participant affordances for the org dashboard — a Message action on session
cards, plus reschedule/cancel and own-documents — are agreed but are feature
work. They follow separately so this correctness fix is not held behind them.

Part of #1021

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected data loading so pending requests, events, and conversations consistently include the appropriate scope.
  * Ensured relevant booking data is not inadvertently hidden when viewing request and event lists.
  * Added safeguards to prevent future scope-related regressions across protected routes.

### DIFF
--- a/__tests__/security/org-scope-defaults.test.ts
+++ b/__tests__/security/org-scope-defaults.test.ts
@@ -1,0 +1,119 @@
+/**
+ * A default scope that silently hides org-funded rows has now been found four
+ * times: the consultee home feed, the consultee payments list, the chat inbox,
+ * and the consultant Requests tab. Each was fixed where it was found; this file
+ * pins the whole class so the fifth one fails a test instead of shipping.
+ *
+ * The failure mode is always the same and always quiet. Nothing errors, no row
+ * is marked missing — a surface simply renders a subset and looks complete. The
+ * Requests case was the worst of the four: `RequestSlotAllocationTab` fetches
+ * the two endpoints an expert uses to allocate slots, both of which drop
+ * org-funded rows when `orgScope` is absent. An org-sponsored subscription was
+ * therefore paid for and then never scheduled, because the request to allocate
+ * its slots never appeared on the only surface that can act on it.
+ *
+ * Why the call sites rather than the defaults: the right default genuinely
+ * differs by surface. A spend report wants personal; an inbox wants everything.
+ * Flipping `resolveOrgScope` to `all` would relocate the bug into the surfaces
+ * this repo spent PR #1029 narrowing. So each caller states its intent, and
+ * these tests make omission loud.
+ */
+
+import { readFileSync } from "fs";
+import { join } from "path";
+import { execFileSync } from "child_process";
+
+const read = (rel: string) => readFileSync(join(process.cwd(), rel), "utf8");
+
+/** ripgrep-free source sweep over the app, restricted to tracked files. */
+function trackedFiles(...globs: string[]): string[] {
+  const out = execFileSync("git", ["ls-files", ...globs], {
+    encoding: "utf8",
+    cwd: process.cwd(),
+  });
+  return out.split("\n").filter(Boolean);
+}
+
+describe("useOrgScope callers state their default", () => {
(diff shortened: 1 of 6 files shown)


teetangh:
Closing in favour of a correct implementation.

This PR made the **personal** inbox show org-funded conversations. That contradicts [ADR 19](docs/enterprise/70-design-decisions/19-personal-vs-org-dashboard-split.md), which states that anything sponsored by an organization lives in that organization's dashboard and that the personal scope pins `organizationId: null` on both sides. The merged inbox also has no answer for a member of several orgs — a conversation carries no visible org label, so three orgs produce one unsortable list.

The replacement duplicates the participant surfaces into the org dashboard, scoped by route (`useOrgScope` already pins `org:<orgId>` under `/dashboard/organization/[orgId]/`), and makes DM channel ids org-aware so the split is exact rather than approximate — today `dm-<a>-<b>` is keyed per pair and carries the org tag of whichever booking created it, so one relationship spanning both contexts cannot separate.

Note for whoever picks this up: the slot-allocation hole this PR also fixed is still live on production — an org-sponsored subscription is paid for and never scheduled, because the only Requests surface in the product sits in the B2C consultant tree and filters org rows out. That is being carried into the replacement work rather than dropped.

PR #1041, opened 2026-07-28
ci: catch enum drift before it surfaces as an unrelated Prisma error

teetangh:
## What happened

`183d0e72` (12 July) deliberately removed `LEMON_SQUEEZY` and `XFLOW` from
`PaymentGateway`, along with the Lemon Squeezy webhook route. The **187 rows
still holding those values were never migrated**, and nothing noticed for 16
days.

It surfaced today as two things that look unconnected to a July schema edit:

- `cleanup-abandoned-payments` failing **every run since 08:04**
- the back-office payments dashboard returning **nothing at all** —
  `getOperatorStats` groups by `paymentGateway` inside a `Promise.all`, so one
  unreadable row took down the whole stats read, not just the gateway breakdown
- the same error in Netlify build logs, though the builds themselves stayed
  green

Prisma's message names the value, not the cause — *"Value 'LEMON_SQUEEZY' not
found in enum 'PaymentGateway'"* — raised wherever a query happens to touch such
a row. A long way from the commit that caused it.

The rows are repointed to `RAZORPAY` separately: 187 across `Payment` (156),
`Refund` (21) and `Dispute` (10). The payout tables were clean. Originals are
captured in `_gateway_drift_backup_20260728`, so it is reversible. **This PR is
the guard, not the repair.**

## What it checks — and what it deliberately doesn't

**FAILS** on rows holding a value the schema doesn't declare. That's the thing
that actually breaks: every read touching them throws.

**WARNS** on labels the database has and the schema doesn't, when no row uses
them. Those are inert — Prisma only throws on *reading a row*.

That distinction matters, and I got it wrong first. My initial version failed on
labels, which would have **red-lit CI on every PR**: the labels are still in the
database right now, because Postgres cannot drop an enum value in place.
Clearing them means recreating the type across all six columns that use it —
a `db push` decision, not something to block the repo on.

Columns are discovered from the catalog rather than hardcoded, so a new
enum-typed column is covered the day it's added. **148 enum columns** on the
current schema.

Skips cleanly without `DATABASE_URL` so forked PRs don't fail a check they can't
run — and the CI step passes the secret explicitly, because a guard that always
skips is worse than none: it looks like coverage.

## Verified both ways

```
against the repaired DB → WARNING on the two inert labels, PASS, 148 columns
STRIPE removed from schema → FAILURE naming all 160 rows across 5 tables:
  Dispute.paymentGateway = STRIPE (10 rows)
  ConsultantPayout.provider = STRIPE (8 rows)
  PayoutAccount.provider = STRIPE (41 rows)
  Payment.paymentGateway = STRIPE (86 rows)
  Refund.paymentGateway = STRIPE (15 rows)
```

`tsc` clean · lint clean. No schema change, so no DB push required for this PR.

## Follow-up worth tracking

The two inert labels remain in the database. Harmless today, but they mean a
stray insert or a restored backup could reintroduce the drift silently. Worth
clearing on the next planned `db push` — the guard will keep warning until then.

Part of #1021

## Summary by CodeRabbit

* **Bug Fixes**
  * Added automated checks to detect mismatches between database enum values and the application schema.
  * CI now flags records containing unsupported enum values, helping prevent database inconsistencies from reaching production.

* **Chores**
  * The check safely skips when database access is unavailable, such as in external pull requests.


teetangh:
Closing in favour of #1042, which covers the same ground with a superset of this check. Not a criticism of the diagnosis here — it identified exactly the right problem, and #1042's `known-drift.json` cites the same root cause (`183d0e72` removed the labels; `db push` cannot drop an enum value in place).

Three reasons the other implementation was kept:

1. **It checks both directions.** A label in the schema that the database lacks fails on *write*, and is just as broken as the read-side P2023 this PR targets.
2. **It has an expiry-bounded allowlist.** Neither PR removes `LEMON_SQUEEZY`/`XFLOW` — that needs a type recreation and a column swap. Without an allowlist the guard turns CI red immediately on drift nobody can fix in the same change, which is how a check gets disabled. `prisma/sql/known-drift.json` carries the entry with an owner and an expiry of 2026-09-30, so it stops excusing itself on a date rather than silently forever.
3. **It sits alongside a sidecar check** that asserts every CHECK/EXCLUDE/index/trigger in `prisma/sql` is actually applied — which found that `collaborator_plan_xor` had been declared for months and never applied to the live database.

On the tsx-versus-mjs point from `95d85156`: the repo already runs `npx tsx scripts/ci/check-money-columns.ts` as a CI step, so a TS runner in that position is the existing convention rather than a new dependency. Worth revisiting for the whole `scripts/ci/` set together if the runner is causing trouble.

**The underlying drift is still unfixed by either PR.** Removing those two labels needs a new enum type, a per-column `USING` swap, and a drop — worth its own change. Both guards would have caught it; neither does the surgery.

PR #984, opened 2026-07-11
chore(payments): drop Lemon Squeezy + XFlow, add DODO_PAYMENTS enum (post-MVP), keep Stripe

teetangh:
Executes the gateway-evaluation ADR (Lemon Squeezy + XFlow = REMOVE). **Stripe retained** for possible future use.

- Removed Lemon/XFlow checkout stubs (`lib/payments/index.ts`), webhook routes (`app/api/webhooks/{lemon-squeezy,xflow}`), config/type/seed refs, and doc mentions; plus stray day-pass doc mentions (no Prisma model exists).
- Added `DODO_PAYMENTS` to the payment-gateway enum with a post-MVP/evaluation-pending comment.

**Schema change** (enum value add + two removals) — needs the centralized `prisma db push` at wave end. Type-check via CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1045, opened 2026-07-29
fix(db): drop LEMON_SQUEEZY and XFLOW from the live PaymentGateway enum

teetangh:
Closes the drift that made `reconcile-disputes` and `cleanup-abandoned-payments` fail with `P2023` for a month.

**Already applied to the database.** The DDL ran via Supabase's migration API before this PR was opened, so the code change here — deleting the tolerance entry — is what makes CI assert the state that now exists. Reviewing and merging this does not trigger any further database change.

## Why it was stuck

`183d0e72` removed both labels from `schema.prisma` when the gateways were retired (#984), but Postgres has no `ALTER TYPE … DROP VALUE`, so `db push` left them live. Prisma then refuses to read *any* column typed by an enum it doesn't fully recognise — which surfaces as `P2023`, an error naming neither the enum nor the gateway. That indirection is why it took a month to attribute.

It was then deferred on the grounds that 156 `Payment`, 21 `Refund` and 10 `Dispute` rows still referenced the dead values. That had stopped being true, and the stale note was itself deterring the fix. Zero rows use either value.

## Checked before writing the migration

| | |
|---|---|
| Columns typed by the enum | 6, all plain tables — no views, functions or generated columns |
| Columns with a `DEFAULT` | none, so no drop/re-add dance |
| Rows across those tables | 555 — each `ACCESS EXCLUSIVE` lock is milliseconds |

That last row mattered more than it looks: production and development share one Supabase project, so this was a lock on the live database.

## Verified after

- Live enum reads `STRIPE, RAZORPAY, DODO_PAYMENTS, CARD`, in `schema.prisma` order
- No `PaymentGateway_old` left behind; all six columns on the new type
- Row counts and per-gateway distributions byte-identical (`Payment` still RAZORPAY 240 / STRIPE 86 / CARD 71)
- Prisma reads of `Dispute`, `Payment` and `PayoutAccount` all succeed — the exact reads that raised `P2023`
- `check-db-drift`: **107 schema enums against 107 live types, 0 tolerated** (was 106 with 1 tolerated)

## One-off, not a sidecar

Sidecars under `prisma/sql/` are re-applied to every database because `db push` never creates them. This is the opposite case: `schema.prisma` already declares the correct four values, so a database built from scratch comes out right on its own. Only the existing drifted one needed correcting, hence `prisma/sql/one-off/` with a date on it.

The 2026-09-30 expiry that would have turned CI red is no longer load-bearing.


teetangh:
**Additive-only conflict — needs an explicit decision before merge.**

This PR drops `LEMON_SQUEEZY` and `XFLOW` from the live `PaymentGateway` enum. Even though the body notes the DDL was already applied out-of-band, that is still a **destructive** schema change (enum value removal), which conflicts with the standing additive-only rule for the shared Supabase project.

Context from the 2026-08-01 session: a column rename (`callsPerWeek` / `meetingsPerWeek` → `sessionsPerWeek`) that landed in the DB ahead of the matching client release caused a brief production outage (`PrismaClientKnownRequestError` on missing columns). That is the failure mode this rule exists to prevent.

Recommendation: keep this PR open and blocked until product/ops explicitly accepts enum-value removal (or rewrite it as additive — e.g. stop writing those values in app code while leaving the enum members in place). Tracked in #1089.

Issue #1072, opened 2026-08-01
Launch readiness master tracker — every open issue banded pre-MVP / post-MVP / scale, grouped by subsystem


teetangh:
> **This is the live launch-readiness tracker.** The two previous ones — #331 (MVP Launch Roadmap) and #328 (MVP Pending Items) — were both closed in March 2026, and nothing replaced them. Since then the backlog has grown to **107 open issues**, of which **44 carry no label at all**, so there has been no single place that answers "what is actually left before we can take money from strangers".

This issue does three things. It sorts every open issue into one of three launch bands, it groups those bands by subsystem, and it records the issues that are stale, wrong, duplicated or in the wrong band so they can be fixed rather than silently rot.

Everything below was checked against `dev` at `839d359c` on 2026-08-01. Where a claim is a code-verified fact rather than a restatement of the issue text, it is marked **[verified]**.

---

## 1. How the bands are defined

The three bands are not a priority ranking. They answer three different questions.

**Pre-MVP** — the platform is about to take real money from people who are not us. An issue belongs here if a launch-day user could lose money, lose data, be unable to finish a booking, or if the failure would happen without us finding out. Regulatory exposure that attaches on the first transaction also belongs here. This band is the launch gate.

**Post-MVP** — the first ninety days after launch. The platform works, but coverage, polish and operational maturity are thin. Nothing here stops us from opening the doors; all of it will start to hurt once real users are inside.

**Scale** — needed only when volume, enterprise demand, or a new revenue line forces the question. Building any of it before launch is speculative work against unvalidated demand.

## 2. Where we are, in one table

| Band | Open issues | Label | What it means |
|---|---|---|---|
| **Pre-MVP** | 30 | `launch: pre-mvp` | The launch gate. |
| **Post-MVP** | 48 | `launch: post-mvp` | First ninety days. |
| **Scale** | 13 | `launch: scale` | Deliberately deferred. |
| **Unbanded** | 13 | — | Carry no band label at all — see the correction below. |
| **Stale / needs triage** | 3 of 12 | — | §6's list, mostly worked through. |

[...]

Issue #997, opened 2026-07-17
Request Calendar performance: slow PENDING list endpoints + move client-side allocation engine server-side


teetangh:
## Summary

The consultant **Request Calendar** (`/dashboard/consultant/[id]/(features)/requests`) has two *distinct* performance problems that were conflated under "the calendar hangs." Investigation (code audit + live production reproduction) separates them:

- **Problem A — slow PENDING-requests list endpoints (server-side).** This is what actually makes the *list page* hang. For a consultant with a pending item, `GET /api/bookings/subscriptions?status=PENDING` sat un-resolved for **30–40s** on production before the page rendered.
- **Problem B — heavy client-side allocation engine (client-side).** The "Allocate Slots" dialog pulls a ~4,600-LOC scheduling/availability/auto-allocation engine into the browser. This is what trips `next dev`'s memory-threshold auto-restart locally and bloats the consultant bundle.

The original hypothesis was "heavy client-side computation." That's correct for **B**, but the evidence shows the user-visible list-page hang is dominated by **A** (a slow server query) — so this needs both a server-query fix *and* a client→server migration, and the migration must be paired with query optimization (naively pushing more work onto an already-slow endpoint would make A worse).

## Reproduction / evidence

- **Local dev:** loading a consultant's Requests page trips `⚠ Server is approaching the used memory threshold, restarting…`, dropping the connection (`ERR_CONNECTION_REFUSED`) so the populated view never renders. `/api/bookings/*` calls returned 200 but at 7–20s each in dev.
- **Production (`familiarisenow.com`):** compiled build, no dev memory watchdog, so it *does* render. Verified across three consultants (seed accounts are live on prod):
  - **David Anderson** (has 1 PENDING) — "Loading requests and availability…" persisted **~40–60s**; `/api/bookings/subscriptions?status=PENDING` stayed `pending` in the network panel the whole time, then resolved and rendered the PENDING "Basic Consultation" card with Use Requested Times / Allocate Slots / Decline.
  - **James Brown / Benjamin Brown** (requests already APPROVED, 0 PENDING) — both `/api/bookings/*` calls returned 200 **promptly**, "No pending requests" rendered with no stall.
- **Signal:** the stall scales with the presence/number of PENDING items needing slot math, and lives in the API response time — i.e. **server-side**, not (only) client render. Screenshots captured under `screens/reqcal/` (`anderson-requests.png`, `brown-james-requests.png`, `brown-benjamin-requests.png`).

[...]

Issue #837, opened 2026-06-10
[UMBRELLA] B2C state-machine and payments hardening — launch blockers + first-month


teetangh:
## Why

The 2026-06-10 staff-level audit (10-agent enterprise pass + 7-agent B2C/chaos/capacity pass) found that the enterprise (B2B) subsystem received the full state-machine hardening pass (PRs #825, #826) while the B2C path — the side real users hit first — did not. Every fix below is a pattern PORT from the enterprise side (idempotency keys, CAS transitions, confirm-time revalidation, limiter coverage), not an invention.

## Launch blockers (PR-C, stacked on #826)

- [x] #788 — slot picker availability-ID mis-binding (silently rejects ~5–15% of legitimate bookings)
- [x] #827 — cross-user tentative-slot double-booking (no confirm-time overlap recheck)
- [x] #828 — checkout request-level idempotency key
- [x] #677 PM-1 — RAZORPAY_KEY_SECRET/RAZORPAY_SECRET env split silently disables all payout/payment reconciliation (verified live on this branch)
- [x] #677 PM-2 — non-deterministic RazorpayX idempotency key (`payout_${id}_${Date.now()}`) can double-disburse on retry

## First-month (PR-D after the chaos suite calibrates frequency)

- [ ] #829 — tentative-cleanup vs webhook confirmation race
- [ ] #830 — orphaned SUCCEEDED-payment / isTentative=true recovery sweep
- [x] #831 — unbounded text fields + missing limiters on consumer mutations
- [x] #832 — appointment lock 60s TTL without renewal
- [x] #833 — 7-day tentative expiration UX lockout
- [x] #834 — waitlist slot assignment outside the checkout tx
- [ ] #835 — org-wallet-funded B2C refunds invisible to org audit (B2B↔B2C seam)
- [x] #836 — subscription approval read-then-decide race
- [ ] #717 Problem 1 — wallet FundingLedgerEntry.currency schema/migration drift (fold into the pre-MVP reset)

## Validation gate

The 8-scenario chaos runbook (landing in docs/enterprise/50-operations/ via PR-C) is the go/no-go: scenarios 1–4 (same-slot race, capacity overrun, double-submit, webhook replay/out-of-order) must pass with zero duplicate charges and zero double-bookings; the 2x-peak load ramp must hold P95 under 2s with no pool exhaustion.

Tracked against the production-readiness verdict of 2026-06-10; full audit artifacts in the session transcript.

[...]

Issue #908, opened 2026-06-21
bug(allocation): subscription Auto-Allocate fails with 500 'Unable to start a transaction in the given time' (~115s) — paid subscription left unallocated


teetangh:
## Summary

Submitting a subscription slot allocation (consultant → Requests → Allocate Slots → **Auto Allocate**) can fail with **HTTP 500** after ~**115 seconds**. The subscription stays `PENDING_ALLOCATION`, so the consultee never receives their sessions. The booking is paid for but cannot be fulfilled.

This is the functional failure mode of the performance problem in #907 — the allocation transaction cannot start because the connection pool is starved by the heavy `availability-with-allocation` computation running (repeatedly/concurrently) around it.

## Evidence (dev server log)

```
{"event":"lock_acquired","key":"auto-allocate:fbe6e6b7-3b85-4747-b37c-bfcdc914641b","attempts":1,"duration_ms":318,"ttl":148500}
{"event":"lock_released","key":"auto-allocate:fbe6e6b7-3b85-4747-b37c-bfcdc914641b","held_duration_ms":33245}
[Subscription Allocation] Failed after 35338ms: Transaction API error: Unable to start a transaction in the given time.
PATCH /api/bookings/subscriptions/cmqnxpyib0005sfyosp8e3ctr/allocate 500 in 115772ms
```

Client console also shows `429 (Too Many Requests)` (×4) from the calendar's duplicated wide availability requests immediately before.

## Endpoint

`PATCH /api/bookings/subscriptions/[subscriptionId]/allocate` (`app/api/bookings/subscriptions/[subscriptionId]/allocate/route.ts`).

## Reproduction

1. As a consultee, buy a 1-month "Basic Subscription" (4 sessions) with Mock Pay.
2. As that consultant: Requests → Allocate Slots. The calendar takes ~28s+ to load (see #907).
3. Click **Auto Allocate**. After ~30–115s the request returns 500; the dialog closes but the request is still **Pending** and no session appointments are created.

## Root cause

`"Transaction API error: Unable to start a transaction in the given time"` is Prisma failing to acquire a connection / begin the interactive transaction within its wait budget. Contributing factors:

- The allocation flow drives the **O(window) `availability-with-allocation`** computation over the full scheduling period (#907) — ~28s cold for 1 month. The allocation calendar fires it **twice concurrently** (React StrictMode in dev) and the lock is held ~33s.
[...]

PR #1091, opened 2026-08-01
fix(booking): Pre-MVP algorithm integrity — #1071 #1012 #1005

### LINKED ISSUE #834: [B2C][WAITLIST] Slot assignment runs outside the checkout transaction; partial multi-session enrollment on mid-loop failure
## Problem
`lib/waitlist/slot-handler.ts` assigns freed slots outside the booking tx, and multi-session class enrollment can fail mid-loop leaving partial enrollment. No unique constraint on the SlotOfAppointment↔User m:n protects against double-assignment.

## Fix shape
Move assignment into the same tx as the slot release (or CAS-claim per slot); add the missing unique constraint at the pre-MVP schema reset; make multi-session enrollment all-or-nothing.

### LINKED ISSUE #1003: Group-event cancellation refunds and notifies nobody (audit C3/M2)
## Context

The 2026-07-17 booking-lifecycle audit found that cancelling a whole webinar or class refunds nobody and notifies nobody. The cancel route only assembles notification recipients for exclusive (1:1) types, its refund block is exclusive-types-only (`app/api/appointments/[appointmentId]/cancel/route.ts:167-187, 306-315, 372-376`), and the participant-removal endpoints contain no payment handling at all (`app/api/participants/webinar/[webinarId]/route.ts:188-211`). Deletion of a paid group event is blocked with "Cancel or refund first", which is circular because no in-app path can refund the attendees. Attendee `ConsultantEarnings` are also never reversed, so a consultant who cancels a webinar keeps the per-attendee earnings.

## Proposed approach

Once the refund-gateway wiring PR lands, iterate the group event's paid attendee appointments on wrapper cancellation: create a gateway refund per attendee payment under the existing policy snapshot, reverse the matching earnings rows through the standard cascade, and fan out cancellation notifications to every enrolled attendee. The attendee-removal endpoint should reuse the same per-attendee unit.

Blocked by the refund-gateway PR (audit M1). Audit ids C3/M2; found alongside #998.

### LINKED ISSUE #1004: Auto-refund on consultant REJECT of a paid request (audit M3)
## Context

When a consultant REJECTs a paid consultation or subscription request, no refund is proposed or created — the word "refund" does not appear anywhere in the status-PATCH handlers (`app/api/bookings/consultations/[consultationId]/route.ts:427-717`; subscriptions equivalent). The consultee has paid, the consultant has declined to serve them, and the money silently stays captured until someone notices in the dashboard.

## Proposed approach

On a REJECT transition of a request whose payment is SUCCEEDED, automatically drive a 100% refund through `refundPayment` (mechanical once the refund-gateway PR wires the actual gateway call) and notify the consultee. The cancellation-policy snapshot is irrelevant here — rejection is the consultant's act, so the full amount applies.

Blocked by the refund-gateway PR (audit M1). Audit id M3; found alongside #998.

### LINKED ISSUE #1005: Consultee self-leave for group events + UI kind-gates for impossible actions (audit R4/R5/C4/C6)
## Context

Several consultee-facing actions are offered by the UI but always fail server-side. Consultees see Reschedule on webinar/class cards although the route 403s them (`app/api/appointments/[appointmentId]/reschedule/route.ts:141-152`); trial reschedule is offered but the participant check never matches TRIAL (`ConsulteeAppointmentsAdapter.tsx:188-200`); "Cancel booking" is shown for group events with no self-removal endpoint in existence (C4); and the consultee trial Cancel calls the generic route, gets a 409, and the toast falsely reports "Booking already updated" (`useEventActions.ts:158-177`). Separately there is no self-leave flow at all for webinar/class attendees — the only removal path is consultant-initiated.

## Proposed approach

Two parts. First, kind-gate the UI: hide or disable actions whose server route will 403 or no-op for the derived appointment type, and point trial cancellation at the trial route. Second, design and build a consultee self-leave endpoint for group events (per-attendee appointment cancellation with policy-snapshot refund once the refund gateway lands), which then lets the UI re-enable the action honestly.

Audit ids R4/R5/C4/C6; found alongside #998.

### LINKED ISSUE #1012: Stale-tab reschedule can delete+recreate a completed allocation — needs allocation precondition (audit R2, design with #997)
## Context

Reschedule flows intentionally omit the `initialAllocation` guard introduced in #998 (they legitimately re-allocate), which leaves one residual race: a stale second tab submitting a full reschedule after the first tab completed it finds zero tentative slots, is classified as an initial allocation, and the manual path delete-and-recreates the winner's completed allocation (`SlotAllocationService.ts:842-875, 2246-2299`). The full-delete branch also lacks the imminent-slot protection the in-progress branch has. Partial-subscription stale tabs fail benignly on the exact-count check; consultations, webinars, and full reschedules are exposed.

## Proposed approach

Add an allocation precondition for non-initial allocations — for example an `expectedTentativeSlotCount` or an `updatedAt` compare-and-set carried from the page that initiated the reschedule — so a stale tab mismatches and receives a conflict instead of replacing state. The natural place to build this is the #997 server-side allocation migration, which owns the request shape; this issue should ride that design rather than a standalone patch.

Audit id R2; found alongside #998. Part of the #997 design.

### LINKED ISSUE #1071: Rescheduling a webinar longer than 30 minutes moves only its first slot, splitting one appointment across two days
Rescheduling a webinar longer than 30 minutes moves only its first 30-minute slot. The remaining rows stay at the original time, so one appointment ends up holding two sessions on two different days.

Found while auditing whether an `Appointment` can ever own more than one session (context: #1061, #1070).

## Mechanism

A webinar allocated through `app/api/bookings/webinars/[webinarId]/allocate/route.ts:85-88` gets `ceil(durationInHours / 0.5)` contiguous 30-minute rows on a single appointment — a two-hour webinar is four rows.

The planner's reschedule path then updates exactly one of them:

```ts
// app/api/bookings/webinars/crud-with-plan/route.ts:777
appointment.slotsOfAppointment[0]
```

Row 0 moves to the new date. Rows 1 to 3 are left where they were.

Two things make it worse:

- The `include` that loads those slots (`:742-750`) has **no `orderBy`**, so which row is `[0]` is not deterministic. The row that moves may not even be the earliest one.
- The `activePayments > 0` guard at `:573-603` only protects webinars that already have paid registrations, so an unsold webinar is rescheduled straight through this path.

## What a user sees

A consultant reschedules their two-hour webinar from Tuesday to Thursday. The listing shows Thursday. Attendees who join on Thursday get the first half hour and then the room stops being valid; the remaining ninety minutes are still sitting on Tuesday, in the past. Depending on which row `[0]` resolved to, the visible start time may be wrong as well.

## Why it matters beyond the obvious

This is currently the only production writer that can put two non-contiguous runs on one appointment. That has a second consequence: it is part of the reason the video room in #1067 must be keyed to a contiguous run rather than simply to `appointment-<id>`. If the room were keyed by appointment, this bug would merge two separate sessions into one shared video room — a cross-session privacy leak rather than a scheduling glitch. See #1070 for that analysis.

## Fix

Update every live slot on the appointment, not `[0]`, preserving the contiguous block: recompute the full run from the new start using the same 30-minute chunking the allocator uses, rather than shifting one row. Add an `orderBy: { startsAt: "asc" }` to the include regardless, since `[0]` being arbitrary is a latent hazard on its own.

Worth pairing with a validator asserting that an appointment's live slots form exactly one contiguous run, which would have caught this at write time.

## Related

- #1061 — the video-room split this contributes to
- #1070 — Stream scheduled-call model, and why room identity cannot rely on `appointmentId` while this exists

teetangh:
## Summary
- Canonical N×30min contiguous slot runs for webinar/class planner CRUD (#1071)
- `expectedTentativeSlotCount` stale-tab reschedule precondition (#1012)
- Consultee kind-gates, trial cancel via trial API, group self-leave (#1005)
- #997 Phase 3: weekly-limit guard uses server `weeklyConfirmedCallCounts` (Phase 0–2 already on `dev`)
- Evidence closes for already-fixed money paths (#1003, #1004) and obsolete waitlist (#834)

Closes #1071
Closes #1012
Closes #1005
Closes #1003
Closes #1004
Closes #834
Part of #676
Part of #837
Part of #997
Part of #1072

## Test plan
- [x] `__tests__/booking-algorithm/contiguous-slot-run.test.ts`
- [x] `__tests__/booking-algorithm/expected-tentative-count.test.ts`
- [x] `__tests__/booking-algorithm/consultee-affordances.test.ts`
- [x] `__tests__/booking-algorithm/server-status-grid.test.ts`
- [ ] Planner: create 2h webinar → 4 slot rows; reschedule → all four move as one run
- [ ] Reschedule allocate from two tabs → second gets 409 after first completes
- [ ] Consultee webinar card: Leave event (no Reschedule/Cancel booking); trial Cancel hits `/api/trials`
- [ ] Existing cancel/reject refund suites still green in CI

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Consultees can leave eligible classes and webinars directly, with clearer confirmation dialogs.
  * Rescheduling is available for consultations and subscriptions, with actions tailored by appointment type.
  * Class and webinar bookings now use reliable contiguous time slots.
* **Bug Fixes**
  * Prevented stale rescheduling pages from overwriting newer booking changes.
  * Improved weekly subscription-limit calculations and slot-update participant preservation.
  * Added attendee-aware refund handling and improved notification display.
  * Enforced 30-minute increments for class durations.
* **Documentation**
  * Updated booking, rescheduling, cancellation, and troubleshooting guidance.


teetangh:
## Review — #1071 / #1012 / #1005

Read the full diff and traced each change against the surrounding code. The three new suites pass locally (13/13). The direction is right — canonicalising planner CRUD onto the allocator's N×30min shape is the correct fix for #1071, and the `expectedTentativeSlotCount` precondition sits under the Redis locks where it belongs. Findings below, worst first.

---

### 1. `replaceContiguousSlotRun` hard-deletes slots, which cascades `MeetingSession` and `Recording` away — Critical

`lib/appointments/contiguous-slot-run.ts:117` does `deleteMany` + recreate. Both children cascade:

- `prisma/schema.prisma:3923` — `MeetingSession.slotOfAppointment … onDelete: Cascade`
- `prisma/schema.prisma:4036` — `Recording.meetingSession … onDelete: Cascade`

The code this replaces did `tx.slotOfAppointment.update({ where: { id: slot.id } })`, which preserved the slot id and therefore the Stream session and every recording row hanging off it.

The `activePayments > 0` guard narrows this (a paid webinar 400s before reaching the transaction), but it does not close it: a free webinar, or one whose payments all landed in `FAILED`/`EXPIRED`, still reaches `replaceContiguousSlotRun`, and a `MeetingSession` exists as soon as the host opens the room once (`getOrCreateAppointmentMeeting`). A duration-only edit is enough to trigger it.

Suggested shape: reconcile rather than replace — update the first `min(existing, N)` live rows in place, create/delete only the delta. Failing that, refuse the rewrite (or detach) when any live slot carries a `meetingSession`.

### 2. Duration-only edits can snap a webinar back to a dead slot's time — High

`app/api/bookings/webinars/crud-with-plan/route.ts:571`

```ts
const existingSlot = webinarToUpdate.appointment.slotsOfAppointment[0];
startTime = existingSlot.startsAt;
```

The PR added `orderBy: { startsAt: "asc" }` to the include, but the include has no `where` — `CANCELLED`/`RESCHEDULED` rows are still in the array. `app/api/appointments/[appointmentId]/reschedule/route.ts:348` stamps replaced slots `{ isTentative: true, completionStatus: "RESCHEDULED" }` and leaves them in place, so on any webinar that has been rescheduled forward, `[0]` is now *deterministically* the old, earlier, dead slot. A subsequent duration-only edit rewrites the live run back to the cancelled time.

The `timeChanged` guard immediately below (`runStart = existingSlots[0]?.startsAt`, `runEnd = existingSlots.at(-1)?.endsAt`) reads the same unfiltered array and has the same problem in both directions.

Filter to live rows (`completionStatus: { notIn: ["CANCELLED", "RESCHEDULED"] }`, and `deletedAt: null` — see #8) in the include, or run the array through `isDeadSlot` before indexing.

### 3. Consultee self-leave issues a full consultant-initiated refund, with no notice window and no past-event check — High (money)

`app/api/participants/{webinar,class}/[id]/route.ts` now lets a consultee `DELETE` themselves, and the handler unconditionally calls `refundRemovedAttendeeSeat`. That helper computes:

```ts
computeRefundPct(policy, /* hoursUntilStart */ -1, /* isConsultantInitiated */ true)
```

and `lib/payments/operations/cancellation-policy.ts:64` short-circuits on `isConsultantInitiated` — it returns `policy.consultantInitiatedPct` and never looks at the tiers. That was correct while only organisers could reach the route. It is not correct for self-service: the consultee gets the organiser-fault percentage regardless of notice, the route has no `startsAt` or status guard, and the refund reason is literally written as `"removed from webinar … by the organiser"`.

Concretely: attend the full webinar, hit **Leave event**, get the money back.

The dialog copy ("a refund is issued under the event's cancellation policy") describes the behaviour we want, not the behaviour we have.

Fix: thread `initiatedBy: "attendee" | "organiser"` into `refundRemovedAttendeeSeat`; on the attendee branch pass the real `hoursUntilStart` so the tiers apply, and reject self-leave once the first live slot has started.

### 4. "Leave event" and "Cancel trial" never render on the appointment detail page — High

`ConsulteeAppointmentsAdapter.tsx` gates both new branches on `sourceId(vm)`, which reads `vm.raw.source.id`. That holds on the list page (`map-consultee.ts` sets `source: w` / `source: c` / `source: t`), but the same adapter is mounted by all three detail clients, and `lib/appointments/map-detail.ts:187` sets `source: detail` where `detail` is `TAppointmentDetail` = `{ appointment, siblings }` — no `id` field.

So `sourceId` returns `null` and neither `else if` fires. On `/appointments/[appointmentId]` a webinar, class or trial now offers **no destructive action at all**; previously it offered Cancel. That's a silent regression rather than the intended honest gating.

Derive the event id from `vm.raw.appointment` (`webinarId` / `classId` / `trialSession.id`) or add an explicit `eventId` to `AppointmentVM` so both mappers are forced to supply it.

### 5. Class durations that aren't 30-minute multiples get silently inflated — Medium

`buildContiguousSlotAtoms` uses `SlotCalculationService.getSlotsPerCall` = `Math.ceil(h / 0.5)`. `WebinarPlanSchema` refines duration to 30-minute increments (`schemas/plans.ts:516`), but `ClassPlanSchema.sessionDurationInHours` (`schemas/plans.ts:547`) only enforces `min(0.5).max(4)`. A 0.75h class previously wrote one 45-minute row; it now writes 2×30 = 60 minutes of the consultant's calendar.

It also breaks the webinar `timeChanged` comparison for such durations — `runEnd` (ceil-rounded) can never equal `endTime` (`start + duration`), so every edit on a paid event would 400.

Either add the increment refine to `ClassPlanSchema`, or clamp the final atom's `endsAt` to `startsAt + durationInHours`.

### 6. The riskiest function in the PR has no test coverage — Medium

Coverage from the new suites: `contiguous-slot-run.ts` lines **108–180 uncovered** — i.e. all of `replaceContiguousSlotRun`. Only the pure builder and the assert are exercised. The delete-and-recreate ordering, user preservation, dead-row handling and the post-write assert are exactly the parts where #1 and #8 live. Worth a fake-`tx` unit test at minimum.

### 7. `expectedTentativeSlotCount` rides on a caller-dependent slot array — Medium

`SlotPicker.tsx:225` derives the count from `subject.slots.filter(s => s.isTentative)`. That's correct for `AllocateClient`, whose `subject.slots` comes from `readAllocationRequest` unfiltered. It is *not* correct for the other two subject builders: `buildRescheduleSubject` and `buildManageTimingsSubject` both use `liveFutureSlots()`, which drops `completionStatus === "RESCHEDULED"` — precisely the rows the reschedule route marks `isTentative: true`.

Today that's latent: the reschedule policies run in `calendarMode: "select"` (no allocate), and `ManageTimingsClient` runs in `"allocate"` mode but never sets `hasReleasedSlots`, so the field stays `undefined` and the guard silently never fires there. The moment anyone sets `hasReleasedSlots` on a reschedule/timings subject, every allocate 409s permanently with "Reschedule state changed in another session".

Two things worth doing: give Manage Timings the guard it currently doesn't have, and make the count come from something with one definition (a server-supplied field on the subject) rather than from a slot array whose filtering differs per builder.

### 8. `isDeadSlot` ignores the A10 tombstone — Low

`lib/appointments/slots.ts:53` checks only `completionStatus`. `SlotOfAppointment.deletedAt` (`schema.prisma:3875`) is a real soft-delete marker. In `replaceContiguousSlotRun` a tombstoned row therefore counts as live: its users get re-attached to the new run, and the `notIn` `deleteMany` hard-deletes the tombstone. Add `deletedAt: null` to the live filter (and to the `deleteMany` `where`).

### 9. Kind-gates still offer two actions the server rejects — Low

The stated goal is "only offer actions the server will honour", but `!inactive` doesn't cover:

- **COMPLETED trial** — `validTransitions.COMPLETED = ["CONVERTED"]`, so "Cancel trial" returns 400 `Cannot transition from COMPLETED to CANCELLED`.
- **Past webinar/class** — "Leave event" returns 200 and refunds (see #3).

### 10. N× writes per checkout — Low (perf)

`handleWebinarCheckout` / `handleClassCheckout` connect the buyer with one `slotOfAppointment.update` per slot, inside the checkout transaction. A 2h webinar goes 1 → 4 updates; a 12-session 2h class goes 12 → 48. `replaceContiguousSlotRun` similarly does N sequential `create`s. `createMany` plus a single connect pass would keep these transactions short.

### 11. The `.map()` in both `crud-with-plan` POSTs is dead code — Nit

```ts
buildContiguousSlotAtoms({...}).map(
  ({ startsAt, endsAt, isTentative, consultantProfileId: ownerId }) => ({
    startsAt, endsAt, isTentative, consultantProfileId: ownerId,
  }),
)
```

This is an identity map — the rename is to the same key. Its only effect is dropping `user`, which is never present because `userIds` isn't passed. ~30 lines in two files that can just be `buildContiguousSlotAtoms({...})`.

---

**Docs note.** `docs/booking/03-slot-math-and-calculations.md` now says the planner path "always writes a contiguous N×30min run", but the class `crud-with-plan` PATCH doesn't touch slot times at all — a class whose `sessionDurationInHours` changes keeps its old run length, unlike a webinar. Worth either handling it or saying so explicitly, since the sentence currently reads as covering both.

**Blocking for me:** #1, #2, #3, #4. The rest can follow up.

teetangh:
## Review round 2 — `30ba4d7` (review-blocker fixes)

Went through the fix commit against the original findings. The four blockers are addressed at the design level and the reasoning in the new doc comments is right: reconciling in place instead of delete+recreate is the correct answer to the cascade problem, filtering dead rows out of `LIVE_SLOTS_INCLUDE` closes the snap-back, `initiatedBy` fixes the refund tier, and the `sourceId` FK fallback does restore the detail page (verified — `readAppointmentDetail` uses `include` throughout, so the scalar FKs and `trialSession.id` are all present).

Not saturated yet, though. Three blockers below, one of which stops the branch building at all.

---

### 1. The branch does not compile — 17 `tsc` errors, all from this commit — Blocking

```
npm run db:generate && rm -f tsconfig.tsbuildinfo && npx tsc --noEmit -p tsconfig.json
```

(the regenerate matters — a stale client adds unrelated noise). Two root causes:

**(a) `{ completionStatus: null }` — 5 sites.** `SlotOfAppointment.completionStatus` is `SlotCompletionStatus @default(SCHEDULED)`, i.e. NOT NULL, so Prisma's filter type is `EnumSlotCompletionStatusFilter | SlotCompletionStatus` with no `| null`:

```
app/api/participants/class/[classId]/route.ts(174,15):   error TS2322: Type 'null' is not assignable …
app/api/participants/webinar/[webinarId]/route.ts(171,15): error TS2322: …
lib/appointments/contiguous-slot-run.ts(218,11):         error TS2322: …
lib/payments/operations/event-refunds.ts(262,15):        error TS2322: …
app/api/bookings/webinars/crud-with-plan/route.ts(490,17)
```

The comment (`SQL NOT IN (...) excludes NULL completionStatus, so we OR-null explicitly`) states a true general fact that doesn't apply to this column — there is no NULL row to exclude. A plain `completionStatus: { notIn: ["CANCELLED", "RESCHEDULED"] }` is both correct and sufficient.

**(b) `LIVE_SLOTS_INCLUDE` loses its literal types.** Hoisting it into a `const` with no contextual type widens `notIn: ["CANCELLED", "RESCHEDULED"]` to `string[]` instead of `SlotCompletionStatus[]`. An inline object literal is contextually typed by Prisma's arg type; a hoisted one is not. That single error poisons the inferred type of `webinarToUpdate` / `existingPlan` and produces **11 further errors in the same file** that look unrelated:

```
route.ts(507,21): Property 'consultantProfile' does not exist …
route.ts(528,22): Property 'webinars' does not exist …
route.ts(648,60): Property 'topics' does not exist …
route.ts(583,10): Parameter 's' implicitly has an 'any' type.
```

Verified end to end: dropping the null branch at all five sites and adding `satisfies Prisma.Appointment$slotsOfAppointmentArgs` to `LIVE_SLOTS_INCLUDE` takes the whole project to **zero** `tsc` errors. Nothing else in the commit is wrong type-wise. (`eslint` on the changed files is already clean.)

### 2. The reconcile loop can collide with itself against `slot_no_confirmed_overlap` — Blocking

`prisma/sql/check-constraints.sql:68`:

```sql
ALTER TABLE "SlotOfAppointment" ADD CONSTRAINT "slot_no_confirmed_overlap"
  EXCLUDE USING gist ("consultantProfileId" WITH =, tstzrange("startsAt","endsAt") WITH &&)
  WHERE ("consultantProfileId" IS NOT NULL AND NOT "isTentative");
```

Not `DEFERRABLE`, so it is enforced at the end of every statement. `replaceContiguousSlotRun` now issues one `update` per row while the siblings still hold their old times, and planner webinar rows sit squarely in the constrained set (`isTentative: false`, `consultantProfileId` set from `existingPlan`).

Move a 2 h webinar from 10:00 to 11:00 — targets `11:00 / 11:30 / 12:00 / 12:30`:

| step | statement | state |
| --- | --- | --- |
| i=0 | `UPDATE s0 SET startsAt=11:00, endsAt=11:30` | `s2` still occupies `[11:00,11:30)`, same consultant, both non-tentative → **23P01** |

`isExclusionViolation` catches it and returns **409 "That time conflicts with another confirmed session on your calendar."** The webinar collides with itself, and the message blames a booking that doesn't exist.

Reachable on any shift smaller than the run length when `activePayments === 0` — a free webinar, or one not yet sold. "Push it an hour later" is an ordinary planner action. Shift-and-shrink is worse: the surplus rows are only retired *after* the update loop, so they're still live non-tentative while the loop runs.

Neither the pre-#1071 single-row `[0]` update nor the interim delete+recreate could hit this — it is specific to the new shape.

Cheapest robust fix is two-phase: one `updateMany` flipping every live row to `isTentative: true` (drops them out of the partial index), then the per-row writes restoring the real `isTentative`. Since the target atoms are contiguous `[)` ranges they can't conflict with each other, so the second pass is safe in any order. Ordering the writes by direction (descending when moving forward) fixes the pure shift but not shift-plus-shrink; making the constraint `DEFERRABLE INITIALLY DEFERRED` also works but is a wider change.

Worth noting the two new `replaceContiguousSlotRun` tests move to a *different day* and to the *same start* — neither exercises an overlapping shift, and the stub `tx` has no constraint to violate, so this class of bug can't surface there.

### 3. Class self-leave is permanently blocked after the first session — Blocking

Both DELETE gates use the **earliest** live slot of the whole event:

```ts
const earliestLive = await prisma.slotOfAppointment.findFirst({
  where: { appointment: { classId }, deletedAt: null, … },
  orderBy: { startsAt: "asc" },
});
if (earliestLive && earliestLive.startsAt.getTime() <= Date.now()) return 400;
```

Correct for a single-session webinar. For a class it's the wrong row: past sessions are stamped `COMPLETED` / `UNVERIFIED`, and `DEAD_COMPLETION_STATUSES` is only `{CANCELLED, RESCHEDULED}` — so session 1 stays "live" forever. From the moment week 1 starts, every self-leave on a months-long class returns `400 Cannot leave an event that has already started`, while the UI keeps offering **Leave event** the whole time. That's precisely the offer-what-the-server-rejects problem #1005 set out to remove, reintroduced on the other side.

The same `orderBy: asc` query drives `hoursUntilStart` in `refundRemovedAttendeeSeat`, so even with the gate lifted a mid-program leave computes a negative notice and lands at 0 %. The local is already named `nextLive` — make it that: `startsAt: { gte: now }` for the notice window, and test "has the program ended" against the **last** live slot rather than the first.

---

### Non-blocking

4. **`isDeadSlot` gained a `deletedAt` check — right call, wide blast radius.** Every consumer changes behaviour: `groupSlotsIntoRuns`, `getSlotJoinState`, `slotsAllowReschedule`, and the session/run math on both dashboards. It degrades safely (callers that don't select `deletedAt` get `undefined`), but a tombstoned slot that used to render as a joinable session now vanishes. That's outside #1071's stated scope — worth a changelog line so it isn't discovered as a regression.

5. **Retired surplus rows keep their attendee connections.** `countWebinarParticipants` / `countUniqueParticipants` don't filter dead slots, so capacity is now partly counted off rows deliberately excluded from the run. Harmless today (same people are on the live rows), but it stops being harmless the first time a shrink and a roster change interleave.

6. **`ClassPlanSchema` refine also blocks edits to existing plans.** Any class already authored at 0.75 h / 1.25 h now fails validation on its next PATCH, not just on create. Fine given the pre-MVP reset, but it's a silent edit-blocker rather than a create-time one.

7. **Scope creep.** `components/notifications/NotificationInbox.tsx` (Novu popover scroll) has nothing to do with #1071/#1012/#1005 and appears in neither the PR description nor the changelog table. Easy to lose at merge.

8. **`createdCount` still returns `createdLive.length`** — the total live-row count, not how many were created. Only tests read it today; the name will mislead the first real caller.

---

**Verification run:** `jest __tests__/booking-algorithm/ __tests__/payments/attendee-removal-refund.test.ts` → 764 passed / 36 suites. `eslint` on the 10 changed files → clean. `tsc --noEmit` → 17 errors, all traced to #1 above.

**Blocking:** #1, #2, #3.

teetangh:
## Review round 3 — `d5c0bf7`, `43bc1c1b`, `717e55e`

All three round-2 blockers are genuinely fixed, and each one landed with a targeted regression test rather than just a code change. Verification below, then one real finding and three notes. This is a much cleaner round — close to saturation.

### Round-2 blockers: verified fixed

**Typecheck.** `npm run db:generate && rm -f tsconfig.tsbuildinfo && npx tsc --noEmit -p tsconfig.json` → **zero errors**, cold. All five `{ completionStatus: null }` sites are gone and the comments now say the right thing (`completionStatus defaults to SCHEDULED and is never NULL`). `LIVE_SLOTS_INCLUDE` was dissolved into the shared `findLiveEventSlot` helper, which types its own `where` as `Prisma.SlotOfAppointmentWhereInput` — that sidesteps the literal-widening problem rather than papering over it, which is the better fix.

**Exclusion-constraint self-collision.** The two-phase flip in `replaceContiguousSlotRun` is correct. Flipping every live row tentative in one `updateMany` drops them out of the partial index (`WHERE … AND NOT "isTentative"`), so a row being written to its new time cannot collide with a sibling that still holds its old one; and because the target atoms are contiguous `[)` ranges they cannot collide with each other as each is restored to non-tentative. I walked the three orderings — pure forward shift, shift-and-shrink, shift-and-grow — and none can trip 23P01 against itself now. The surplus-retire loop correctly relies on rows already being tentative from the pre-pass. `tentative-flips the whole live run before an overlapping forward shift` covers exactly the case I described.

**Class self-leave.** Right fix, and the asymmetry is now deliberate and documented: webinar keys on the earliest live atom (`order: "asc"`), class on the last (`order: "desc"`), because a months-long class keeps past sessions live for run math. The refund side is separately correct — `startsAtGte: now` makes `hoursUntilStart` the *next* session rather than a past one, so a mid-program leave no longer collapses to 0%. Extracting all three call sites into `lib/appointments/live-event-slot.ts` is the right call; that filter had already drifted once.

Tests: 1003 passing across `booking-algorithm` and `payments`. Lint clean on the changed files — the two `no-explicit-any` warnings in `SlotAllocationService.ts:2504,2557` are pre-existing (blame: 2025-10-03 and 2026-06-15), not from this PR.

---

### 1. The in-transaction re-assert doesn't serialize on the partial-reschedule path — Medium

`43bc1c1b` is a genuine improvement, but it doesn't close the race on the case that matters most, and the reason is subtle enough to be worth spelling out.

`assertExpectedTentativeSlotCountInTx` is a bare `COUNT`-then-compare inside the write transaction, and the allocate transactions specify only `maxWait` / `timeout` — **no `isolationLevel`** (`SlotAllocationService.ts:1015`, `:1429`, `:1691`), so they run at Read Committed. Two concurrent transactions both read the same tentative count, both pass the assert, both proceed to delete-and-recreate.

What saves the *full* reschedule is incidental: when every slot has been released, `existingNonTentativeSlotCount === 0`, so `isFreshAllocation` is true, `guardInitialAllocationInTx` runs, and it takes `pg_advisory_xact_lock` — which serializes the whole transaction per event. The new assert sits after that lock and inherits its protection.

On a **partial** reschedule — a subscription where some sessions are released and others stay confirmed — `existingNonTentativeSlotCount > 0`, so `isFreshAllocation` is false, `guardInitialAllocationInTx` is skipped, **no advisory lock is taken**, and the new assert runs unserialized. That is precisely the release-specific-sessions flow that `SlotPicker`'s partial mode drives, and precisely where `expectedTentativeSlotCount` is most load-bearing.

The neighbouring doc comment already describes this exact failure and its solution:

> Under Read Committed, two concurrent transactions could BOTH count zero confirmed slots before either commits … so the count alone is not atomic. The advisory xact lock serializes the guarded transactions per event.

**Proposed fix.** Take the advisory lock whenever the guard needs to be atomic, not only when the allocation is fresh — either by hoisting the `pg_advisory_xact_lock` line out of `guardInitialAllocationInTx` and running it unconditionally at the top of the write transaction, or by extending the `if` to `isFreshAllocation || expectedTentativeSlotCount !== undefined`. The lock is keyed per event and released at commit, so making it unconditional costs one extra round trip on paths that currently skip it and removes a conditional nobody has to reason about.

The commit message says the check now catches reschedule races "before delete/recreate", which is true for full reschedules and not yet true for partial ones — worth amending so the next reader doesn't over-trust it.

### 2. The in-txn re-read runs inside the pinned transaction — Note

`assertExpectedTentativeSlotCountInTx` issues a `findMany` with `include: { slotsOfAppointment: true }` inside the interactive transaction. `SlotAllocationService.ts:907` is explicit that reads were deliberately moved *out* of the write transaction under `#908`, because an interactive transaction pins its pooled connection for its whole duration and that was producing "Unable to start a transaction in the given time" against Supavisor.

This only fires on reschedules and it is one indexed query, so the impact is small and the correctness benefit is worth it. But if the advisory lock from finding 1 gets added, a `count` on `slotOfAppointment` filtered by `isTentative` would do the same job without loading every slot row through the relation.

### 3. Consultation plans are now the odd one out on 30-minute alignment — Note

Adding the increment refine to `ClassPlanSchema` was right, and the grandfather clause in the class PATCH is a nice touch — dropping an unchanged `sessionDurationInHours` before validation means editing the title on a legacy 0.75h plan no longer 400s. Good instinct.

But `ConsultationPlanSchema.durationInHours` (`schemas/plans.ts:235`) still has only `min(0.5).max(8)` with no alignment refine, while webinars (`:516`) and now classes (`:547`) both have one. Consultations feed the same `getSlotsPerCall` = `ceil(hours / 0.5)`, so a 0.75h consultation still books 60 minutes of the consultant's calendar. That behaviour predates this PR and is not a regression — but the rule is now inconsistent across three plan types that share one slot-expansion function, which is the kind of gap that gets rediscovered as a bug later. Either add the refine or leave a comment on `ConsultationPlanSchema` saying why consultations are exempt.

### 4. The grandfather re-read could ride along with an existing query — Nit

The `classPlan.findUnique` added before validation runs on every class PATCH carrying a duration, and the route fetches the plan again for the ownership check moments later. Authentication does happen first (`getSession()` at the top of `PATCH`), so there is no unauthenticated-read concern — it is purely a redundant round trip that could fold into the fetch that follows.

---

**Verification run:** `tsc --noEmit` cold → 0 errors. `jest __tests__/booking-algorithm/ __tests__/payments/` → 1003 passed / 62 suites. `eslint` on the changed files → 2 warnings, both pre-existing.

**Blocking:** nothing. Finding 1 is worth fixing before merge since it is three lines and the partial-reschedule flow is the one #1012 was filed for, but it is strictly better than `dev` either way.

PR #1123, opened 2026-08-09
fix: stop caching degraded renders, and correct the root cause of the 30s tail

### LINKED ISSUE #1119: Degraded fail-open render gets cached into the durable ISR cache and served to everyone
## What happens

`fallbackOnTransientDbError` rethrows during `next build` (correctly — #932), but it returns the fallback at request time. On an **on-demand ISR route** that fallback is rendered into a `200` response on a cacheable route, so Netlify stores it in the durable cache and serves it to every later visitor for the whole revalidate window.

## Measured

2026-08-09, deploy preview 1118, anonymous, `/explore/experts/[consultantId]`. Across 42 cold on-demand renders, **2 returned `200` carrying the degraded `ConsultantUnavailable` shell**:

```
C id=6469cc88 ttfb=30.827749 bytes=66407  degraded=1
(healthy renders: bytes=104981..113897, degraded=0)
```

The degraded body's only `<h1>` is:

```
This profile is taking a moment to load
```

Re-fetching that URL afterwards served the **same degraded page from cache in ~0.3 s** with a climbing `age`. So the symptom inverts: the broken page is the *fast* one, and there is nothing slow left to notice.

## Why it matters

The blast radius is one profile page per incident, bounded by the revalidate window, but it is silent — `200`, no error boundary, no Sentry error (only a `warning` breadcrumb from `reportTransient`), and the cached copy hides the original slowness.

Fail-open and cacheable responses are each safe alone and unsafe together.

## Options (not yet chosen)

- Rethrow on a cacheable render, not just during the build — i.e. extend the `isProductionBuild()` guard in `lib/data/fail-open.ts` to any prerender/ISR render, so the request 500s and nothing is cached.
- Or keep degrading but mark the response uncacheable, so the fallback is only ever served to the one request that hit the timeout.

Root cause of the underlying timeout is #1120.

Found while decomposing the cold TTFB — see PR #1118.

### LINKED ISSUE #1120: The 3s Prisma connect budget is not enforced — connect timeout fires at t+29.8s and pins the function for 30s
## Claim under test

`lib/prisma.ts:7-37` tunes a 3 s connect budget and a 6 s query budget so that "a stuck pooler can't pin a Netlify function up to its ~10s ceiling". Worst case is documented as 3 + 6 = 9 s.

**Neither bound holds in production.**

## Measured — 2026-08-09, deploy preview 1118, `/explore/experts/[consultantId]`

A single invocation, paired from the function log (an invocation *start* is an info line with an empty message; adding the duration reproduces the end timestamp exactly):

```
12:36:09.340  (invocation start)
12:36:39.120  prisma:error timeout exceeded when trying to connect
12:36:39.273  prisma:error Connection terminated due to connection timeout
12:36:39.454  Duration: 30113.69 ms | Memory Usage: 913 MB
```

The connect error fires at **t+29.78 s** against a configured `connectionTimeoutMillis` of **3,000 ms** — about 10x over. The client saw 32.23 s TTFB for this request.

Confirmed **not** an env override: `PG_CONNECT_TIMEOUT_MS` and `PG_QUERY_TIMEOUT_MS` are unset on production, deploy-preview and branch-deploy (only `PG_POOL_MAX=1` is set).

Confirmed **not** the build-phase branch: `IS_NEXT_BUILD` would also make `fallbackOnTransientDbError` rethrow, but the same class of request rendered the degraded page instead (#1119), which only happens when `isProductionBuild()` is false. So `NEXT_PHASE` is unset at runtime and the 3 s budget is the one selected.

Separately, invocations of **26.4 s – 31.9 s** were logged on a Pro account whose documented synchronous function ceiling is 26 s, so the platform bound is not holding either.

## Shape of the problem

42 cold renders, bimodal with **no sample between 6.8 s and 30.8 s**:

| Concurrency | Instances | n | Result |
|---|---|---|---|
| 1, sequential | new | 8 | 1.80–2.72 s, no outliers |
| 12 concurrent | mostly new | 12 | 6 at 1.90–4.66 s, 6 at 30.99–33.08 s |
| 16 concurrent | 12 warm | 16 | 12 at 2.64–2.94 s, 4 at 30.83–33.12 s |
| 12 concurrent | all warm | 12 | 3.33–5.89 s, 0 slow |

Slow-response count equals the number of *new* instances in each batch. The tail is the first pooler connect on a new instance when several new instances connect at once — not boot (1.9 s sequential) and not concurrency (warm 12-way is clean).

## Why it matters

The 9 s budget is the stated defence against a saturated Supavisor pinning a function. It is not effective, so worst-case latency is currently unbounded in practice, and #1119 turns some of those timeouts into permanently cached broken pages.

## To investigate

- Where the ~30 s actually comes from — `pg` `connectionTimeoutMillis` is not being honoured on this path, or something above it retries. Related: `PG_POOL_MAX=1` (#1117).
- Whether the Netlify function timeout for the Next server handler differs from the documented 26 s.

Instrumentation added in PR #1118 logs the resolved budgets once per instance boot (`[Prisma:INIT]`), so the deployed values are now readable from Sentry rather than inferred from source.

teetangh:
Two halves of the same failure path. The headline is that **#1120's premise was
wrong, and so was mine when I started**: the ~30 s tail is not a database connect,
and no timeout value can bound it. #1119 is real, worse than filed, and fixed.

## #1120 — the connect budget is not enforced because the event loop is stalled

The budget is wired correctly. `@prisma/adapter-pg` 7.7.0 hands its config
straight to `new pg.Pool(...)` (`dist/index.mjs:752`), `connectionTimeoutMillis`
is a real `pg.PoolConfig` key, and `pg` honours it at both levels — pool queue
wait (`pg-pool/index.js:219`) and socket connect (`pg/lib/client.js:148`).
Reproduced locally against a black-holed TCP endpoint with the deployed config:

```
raw pg.Pool: ["timeout exceeded when trying to connect", 3003ms]
             ["timeout exceeded when trying to connect", 3003ms]
             ["Connection terminated due to connection timeout", 3011ms]
```

Those are the exact two error strings from the production log, at 3.00 s, in the
same order. Prisma does not retry either: one query, one attempt, 3.1 s.

So why did production log that timeout at t+29.78 s? **Because both pg timers are
plain `setTimeout`s, and a `setTimeout` cannot fire while the event loop is
blocked.**

A temporary diagnostic route on this preview reported instance identity,
`process.uptime()`, an event-loop lag probe, and per-attempt connect timings in
its response body — the function log strips our `console.*`
(`compiler.removeConsole`) and never exposed instance identity. It runs **400 ms
of pure idle `await` before touching the database at all**. Forty invocations
across two 20-way rounds; three came back like this:

| instance age | `process.uptime()` | invocation | 400 ms idle phase took | max loop lag | first DB attempt |
|---|---|---|---|---|---|
| 40 ms | 3,363 ms | 1 | **23,905 ms** | 23,746 ms | ok in 862 ms |
| 39 ms | 3,361 ms | 1 | **24,634 ms** | 24,534 ms | ok in 866 ms |
| 100 ms | 3,355 ms | 1 | **24,823 ms** | 24,655 ms | ok in 1,037 ms |

The other 37, all on instances at least 47 s old, completed the same idle phase in
400–453 ms with 1–70 ms of lag. The stall is ~24 s, it happens **before any
database work**, and only on a brand-new instance serving its first invocation.
In an earlier round where a connect was caught by the stall, it failed after
25,723 ms and 26,112 ms and the **immediately retried attempt connected in 340 ms
and 327 ms**. The database was never the slow part.

### What this means

- Nothing in `lib/prisma.ts` can fix it. Do not propose tuning
  `PG_CONNECT_TIMEOUT_MS`; the timer it configures cannot run.
- `prisma:error timeout exceeded when trying to connect` in the function log is
  not evidence of an unhealthy pooler. Check instance age first.
- The "3 s + 6 s = 9 s worst case" comment in `lib/prisma.ts` has now misled two
  investigations. It is replaced with the measurement.

### Ruled out along the way, so nobody re-checks

- `NODE_OPTIONS=--max-old-space-size=6144` reaching the function and letting V8
  overrun a 1024 MB container. `process.env.NODE_OPTIONS` is **`null`** inside the
  function; `netlify.toml` `[build.environment]` does not reach the runtime.
- `IS_NEXT_BUILD` being true at runtime and selecting the 30 s build budget. Next
  sets `NEXT_PHASE` only inside `next build` (`next/dist/build/index.js:1002`) and
  does not inline it via DefinePlugin; the diagnostic confirms `NEXT_PHASE` is
  `null` at runtime.
- Prisma or pg retrying internally. Neither does.

**Inferred, not measured**: that the stall is cold-instance JS/GC work at Lambda's
memory-proportional CPU share. The function reports
`AWS_LAMBDA_FUNCTION_MEMORY_SIZE=1024`, a V8 heap limit of 1,018 MB and 675–795 MB
RSS at rest. Raising function memory would raise CPU with it, but Netlify does not
expose that on this plan. Filed as #1124 with the raw numbers rather than guessed
at here.

### So what actually ships for #1120

Only the corrections: the false comment in `lib/prisma.ts`, and the skill. The
~30 s tail is **not** removed and cannot be from application code — see the
numbers below, where it is unchanged. The stall is filed as #1124.

**A request-time retry was written, measured, reviewed and REVERTED inside this
PR** — the same thing #1116 did with its second change. It was the one lever that
looked endorsed by the data: the attempt straight after a stalled connect
connected in 327 ms and 340 ms. But that is two diagnostic samples, and in the
page-level A/B its effect could not be separated from the rethrow, because zero
transient failures were observed on the fixed deploy at all. Meanwhile review
surfaced a structural cost I had not priced: retrying per read doubles the query
count on the pages that issue four of them, `PG_POOL_MAX=1` serialises those, and
against a genuinely saturated pooler that pushes a failing render toward Netlify's
function ceiling — where the response is a bare platform 500 with **no error
boundary and no `Cache-Status`**, strictly worse than the fast 500 the rethrow
already produces. Unproven benefit, structural downside, so it goes. Reinstating
it needs a per-render budget and fault-injected evidence; noted on #1124.

## #1119 — a degraded render was being written to the durable cache

Reproduced first-hand on `dev` at `1fa94c17`, and it is worse than filed. Across
68 concurrent cold renders of `/explore/experts/[consultantId]`, **10 returned
HTTP 200 carrying the degraded shell** at 66 KB against a healthy 104–118 KB, each
with `Cache-Status: "Netlify Durable"; fwd=uri-miss; stored`. Re-fetched five
minutes later:

```
a146657d… 0.639s 66068B degraded=1 age:318 "Netlify Durable"; hit; ttl=-17
8997c4a9… 0.386s 66066B degraded=1 age:350 "Netlify Durable"; hit; ttl=-50
8d7234c7… 0.370s 66066B degraded=1 age:320 "Netlify Durable"; hit; ttl=-20
a36cbedc… 0.301s 66066B degraded=1 age:319 "Netlify Durable"; hit; ttl=-19
a5536448… 0.394s 66066B degraded=1 age:318 "Netlify Durable"; hit; ttl=-18
8f22f5f6… 0.333s 66066B degraded=1 age:321 "Netlify Durable"; hit; ttl=-21
```

Six broken profile pages, served to everyone in 0.30–0.64 s. And note the
**production of** those entries was fast too — six of them rendered in 0.47–2.90 s,
because a pooler that fails fast reaches the fallback fast. The assumption that a
degraded render announces itself by being slow is wrong.

### The mechanism, verified rather than assumed

Both framework behaviours that make rethrowing the right answer were checked on a
temporary ISR route shaped exactly like the real one (`revalidate` +
`generateStaticParams` returning `[]`), whose render could be made to throw:

- **A thrown render caches nothing.** `HTTP 500` with
  `Cache-Status: "Netlify Durable"; fwd=bypass, "Netlify Edge"; fwd=miss; fwd-status=500`
  — no `stored` — and three consecutive requests each re-rendered rather than
  hitting a cache. Contrast the healthy render's `fwd=uri-miss; stored`.
- **A thrown revalidation keeps the last good copy.** On a `revalidate = 10` route
  whose render throws on alternating 30 s wall-clock windows, 45 polls over three
  minutes returned the good body every single time, with `age` climbing to 39–53 s
  during throwing windows and resetting only when a non-throwing window
  regenerated it. This is Next's documented "Handling uncaught exceptions"
  behaviour, now observed here rather than quoted.

### The change

Fail-open is now **opt-in per call site** and defaults to rethrowing, because the
default has to be safe when someone adds a `revalidate` export to a route later:

- `emptyOnTransientDbError` / `fallbackOnTransientDbError` take
  `{ perRequest?: boolean }` and only degrade when it is set.
- The only opt-in is `app/explore/programs/page.tsx`, which is `force-dynamic`, so
  its fallback reaches exactly the one request that hit the timeout.
- `/explore/experts/[consultantId]` no longer renders `ConsultantUnavailable`; the
  component is deleted. It retries once, then lets `error.tsx` handle it.
- `[orgSlug]`'s `generateMetadata` no longer degrades to a generic title, which was
  right while that route was dynamic and wrong since #1110 made it ISR.
- On `/`, `/explore/experts` and `/explore/enterprise/organisations` the wrappers
  were dropped rather than left in place with the flag unset. Keeping a `.catch()`
  that only rethrows would be dead code that reads as if the page still degrades,
  and the comments claiming exactly that are corrected.
- Be honest about what rethrowing costs. On `/` it replaces the **whole** landing
  page with `app/error.tsx`, not just the section that failed. That is acceptable
  only because `/` is prerendered at build, so a cached copy almost always exists
  and a failed revalidation keeps serving it; the exposed window is a regeneration
  straight after a `revalidatePath` purge. On the two `[param]` routes
  `generateStaticParams` returns `[]`, so a parameter rendering for the first time
  has **no** copy to fall back on and its visitor gets the error boundary. Three
  comments overstated this as "the last good copy keeps serving" and are fixed.
- The two Route Handlers (`/api/announcements`, `/api/user/consultants`) are
  untouched — they use `isTransientDbError` directly and are genuinely
  per-request.

### One risk I nearly shipped (in the reverted half)

`withTransientRetry` wraps `React.cache`d readers on both `[param]` routes, and if
`cache` memoized rejections the retry would silently replay the failure instead of
re-running the query — inert, and invisible to every test. Checked against the
React that Next 15.5.15 actually vendors
(`next/dist/compiled/react`, `19.2.0-canary-0bdb9206`) rather than the `react@18.3.1`
in `package.json`:

```
async attempt 1 -> ERR pool timeout   calls=1
async attempt 2 -> ok-2               calls=2
sync  attempt 1 -> ERR boom           c2=1
sync  attempt 2 -> ok2                c2=2
```

`cache` memoizes fulfilled results only. The retry really re-runs. Noted in the
helper so nobody has to re-derive it.

## Numbers

`/explore/experts/[consultantId]`, anonymous, distinct parameters so each is a
real server render, client-side `time_starttransfer`. Order-balanced across the
two deployments. `dev` = branch deploy at `1fa94c17`.

| | n | degraded + cached | HTTP 500 | > 10 s | range |
|---|---|---|---|---|---|
| **dev (before)** | 68 | **10** | 12 | 23 | 0.47–38.21 s |
| **preview (after)** | 64 | **0** | 0 | 15 | 1.58–34.23 s |

Per round, so the bimodality is visible rather than averaged away:

| round | n | fast mode | slow mode | degraded | 500 |
|---|---|---|---|---|---|
| dev, 12 concurrent | 12 | 1 at 1.85 s | 6 at 30.67–32.68 s | 1 | 5 at 36.3 s |
| dev, 12 concurrent | 12 | — | 5 at 32.03–36.17 s | 3 | 7 at 38.1 s |
| dev, 16 concurrent | 16 | 16 at 0.47–5.34 s | — | 6 | 0 |
| dev, 12 + 16 concurrent | 28 | all 2.23–5.05 s | — | 0 | 0 |
| preview, 12 concurrent | 12 | 6 at 2.03–3.18 s | 6 at 28.07–33.08 s | 0 | 0 |
| preview, 16 concurrent | 16 | 13 at 1.58–6.53 s | 3 at 33.54–34.19 s | 0 | 0 |
| preview, 12 + 12 concurrent | 24 | all 2.14–5.26 s | — | 0 | 0 |
| preview, 12 concurrent (final head) | 12 | 6 at 2.11–4.90 s | 6 at 30.85–34.23 s | 0 | 0 |

Sequential control, 8 renders each on new instances: `dev` 1.79–4.94 s, preview
1.85–2.28 s after discarding the first-ever request to a fresh deploy (32.9 s,
which is the same cold-instance stall).

Read this honestly:

- **The ~30 s tail is unchanged.** 6 of 12 slow before, 6 of 12 slow after; 3 of 16
  slow on the preview. That is the expected outcome given the root cause, and it is
  why this PR does not claim to fix it.
- **The cached broken page is gone**: 10 of 68 → 0 of 64, and the preview's 15 slow
  renders — the exact population that degraded on `dev` — all returned healthy
  103–118 KB bodies with HTTP 200. The last round was re-run against the final head
  after the cleanup commit, and all four public pages were smoke-checked on it.
- The 500s are Netlify platform timeouts at 36–38 s (27-byte bodies, no
  `Cache-Status`), not our error boundary. Their absence on the preview is **not**
  claimed as an effect of this change; the two deployments met different pooler
  conditions.
- One preview round was **discarded** as invalid: its ids turned out to overlap an
  earlier round, so it measured cache hits rather than renders.

We never observed the new rethrow path fire under real load — every transient
failure on the preview was recovered by the retry. It is covered by the throw-route
experiment above and by unit tests, not by production traffic.

### Route table on the merged SHA, unchanged

Pulled from the `TypeScript, Tests & Build` check on `c848050b`, because a change
to how these pages handle errors must not quietly change how they render:

```
┌ ○ /                                                    23 kB   272 kB   1h   1y
├ ○ /explore/enterprise/organisations                  11.7 kB   280 kB   5m   1y
├ ● /explore/enterprise/organisations/[orgSlug]          360 B   196 kB
├ ○ /explore/experts                                   23.5 kB   320 kB   5m   1y
├ ● /explore/experts/[consultantId]                    31.7 kB   362 kB
├ ƒ /explore/programs                                  17.7 kB   320 kB
```

Identical to the shape #1110 established. The empty Revalidate column on the two
`●` routes is correct rather than broken — `generateStaticParams` returning `[]`
produces no build-time entries to source it from. Note how cleanly this lines up
with the change: every route that now rethrows is `○` or `●`, and the single route
that still degrades is the only `ƒ` one.

## Review round

CodeRabbit could not review this PR — its account-level rate limit did not clear
across five requests over an hour, and it reports "Review rate limited" rather
than a review. Rather than merge on a skipped gate I ran an independent adversarial
review of the full diff. It found seven real problems; all seven are fixed here.

- **The retry was inconsistent on `[orgSlug]`.** `generateMetadata` got it, the
  page body at line 274 did not, and since `React.cache` does not memoize
  rejections the body could not inherit the retried result. Moot now that the
  retry is reverted, and both call sites are bare and identical.
- **`[consultantId]/error.tsx` was inverted and double-reported.** Its doc comment
  said the transient "is caught upstream and never reaches here" — the exact
  opposite of the new design — and it called `Sentry.captureException`
  unconditionally while `onRequestError` (`instrumentation.ts:13`) already reports
  the same server error. At the measured rate that is two Sentry issues per
  timeout. Now guarded on `error.digest`, which server errors carry and
  client-side ones do not.
- **`sentry.shared.config.ts` justified leaving pooler timeouts unfiltered** on the
  premise that fail-open degrades them and any occurrence means a new unprotected
  route. Five routes are now deliberately "unprotected", so that detector is
  permanently tripped. Comment rewritten; the strings stay unfiltered because they
  are the only signal that #1124 is still happening.
- **A client-side rationale in `OrganisationsInteractiveContent.tsx`** explained
  its `initialItems.length > 0` guard entirely in terms of a degraded empty page
  that can no longer occur. The guard is still right for a different reason, which
  it now states.
- **Three page-header comments** still credited "fail-open's build-phase rethrow"
  for a guard those files no longer import.
- **`EMPTY_EXPERTS_METADATA`** lost its last importer and is removed. ESLint does
  not catch a dead export.
- **The tests pinned the wrong thing.** #1119 was never a bug in the helper's
  default — it was call-site wiring, and nothing in the suite would have failed if
  someone added `{ perRequest: true }` to an ISR route tomorrow. Added
  `__tests__/explore/isr-routes-never-fail-open.test.ts`, which reads the route
  sources and asserts that no route exporting `revalidate` opts into degrading and
  that everything which degrades is `force-dynamic` or a Route Handler.
  Mutation-checked: adding `perRequest` to `/explore/experts` flips two tests red
  and names the file; removing it goes green. It also caught a false positive in
  its own first draft.

Two findings I looked at and left alone, both pre-existing and outside this
change: `getViewerOrgs` and `fetchProgramStats` in `/explore/programs` use blanket
`catch` blocks that swallow every error class rather than the transient one, and
`/api/announcements` degrades without the `no-store` header its sibling
`/api/user/consultants` sets deliberately. Neither is a live bug — that route is
`force-dynamic` and Route Handlers are uncached by default in Next 15 — but both
are the same hazard class, and crucially the new test **cannot see them** because
they never touch the helper. Filed as #1125.

## Also

- The two diagnostic routes are removed in the last commit. They existed only to
  produce the numbers above.
- `.claude/skills/nextjs-netlify-caching/SKILL.md` carried the wrong root cause for
  the tail and is corrected, including the places that referenced it indirectly.
- Verified with scoped `tsc --noEmit`, `eslint` on every touched file, and the full
  Jest suite (229 suites, 2578 tests). `next build` was not run locally.
- No schema change, no `prisma db push`, no test data written to the shared
  database. All measurement traffic was anonymous reads.

Closes #1119

Deliberately **not** `Closes #1120`. That half was diagnosed, not fixed: the ~30 s
tail is unchanged and cannot be removed from application code. #1120 should be
updated with the root cause and closed by hand, or left open against #1124, which
tracks the stall itself. Refs #1120, #1124.


teetangh:
@coderabbitai review

teetangh:
@coderabbitai review

teetangh:
@coderabbitai review

teetangh:
@coderabbitai review

teetangh:
@coderabbitai review

teetangh:
@coderabbitai review

teetangh:
@coderabbitai review

teetangh:
@coderabbitai review

teetangh:
@coderabbitai review

Issue #1125, opened 2026-08-09
Bare catch-and-placeholder blocks are the same hazard as fail-open, and the new #1119 guard cannot see them


teetangh:
## Context

#1123 made fail-open opt-in, so degrading and being cacheable can no longer coexist through `lib/data/fail-open.ts`, and `__tests__/explore/isr-routes-never-fail-open.test.ts` now pins that invariant across every route file.

That guard has a blind spot, found while reviewing #1123: it only sees degradation that goes **through the helper**. A bare `catch` that swallows an error and returns a placeholder is the same hazard and is invisible to both the helper and the test.

## The three found so far

`app/explore/programs/page.tsx` has two, and they are broader than the helper ever was — they swallow **every** error class, not just the transient one, which directly contradicts the contract the helper documents ("real defects still surface"):

```ts
// lines 29-31
catch { return {}; }        // getViewerOrgs
// lines 139-141
catch { return null; }      // fetchProgramStats
```

Both are safe **today** purely because that route exports `dynamic = "force-dynamic"`, so the degraded response reaches exactly the one request that hit the failure. The moment anyone gives that route a `revalidate` export, they are #1119 again — and no audit of `perRequest` will find it, because these never touch the helper.

`app/api/announcements/route.ts:62` returns its degraded branch as `NextResponse.json({ success: true, data: [] })` with no cache header, while the sibling `app/api/user/consultants/route.ts:210-214` deliberately attaches `Cache-Control: no-store` to its degraded branch with a comment explaining why. Route Handlers are uncached by default in Next 15 and the degrade sits outside the `unstable_cache`, so I do not believe this is a live bug — but the asymmetry is unexplained, and one of the two is wrong about whether the header is needed.

## Why it is worth doing rather than filing and forgetting

The #1119 blast radius was not obvious from reading the code, which is exactly why it survived review twice: a degraded render is a `200`, it is often the *fastest* response on the page (measured at 0.47 s), and once cached it hides the slowness that caused it. Any swallow-and-placeholder on a cacheable route reproduces that shape.

## Suggested work

1. Sweep for bare `catch` blocks that return a placeholder in Server Components and Route Handlers, and decide per site: rethrow, route through the fail-open helper with an explicit `perRequest`, or document why the swallow is correct.
2. Narrow the two in `/explore/programs` to the transient class so a real defect surfaces instead of rendering an empty programs page.
[...]

Issue #1124, opened 2026-08-09
A newly created function instance stalls its event loop for ~24s before any application work


teetangh:
## What was measured

A newly created Netlify function instance stalls its **event loop** for roughly 24 seconds on its first invocation, before any application work. This is the whole of the ~30 s tail on public routes, and it is what made the Prisma connect budget look broken in #1120 — a `setTimeout` cannot fire on a blocked loop, so pg's 3 s connect timer reported at ~26 s.

Measured on deploy preview 1123 with a temporary diagnostic route that ran **400 ms of pure idle `await`, touching no database at all**, while sampling event-loop lag every 50 ms. Forty invocations across two 20-way concurrent rounds:

| instance age | `process.uptime()` | invocation | 400 ms idle phase took | max loop lag | first DB attempt |
|---|---|---|---|---|---|
| 40 ms | 3,363 ms | 1 | **23,905 ms** | 23,746 ms | ok in 862 ms |
| 39 ms | 3,361 ms | 1 | **24,634 ms** | 24,534 ms | ok in 866 ms |
| 100 ms | 3,355 ms | 1 | **24,823 ms** | 24,655 ms | ok in 1,037 ms |

The remaining 37, every one of them on an instance at least 47 s old, completed the same idle phase in 400–453 ms with 1–70 ms of lag. So the stall is bimodal, it is confined to brand-new instances serving invocation 1, and the database is fast on both sides of it.

## Why it matters

It is the only remaining cause of 30-second page loads on the public routes, and it scales with instance creation, which means it fires exactly when traffic arrives in a burst. It cannot be mitigated from application code: #1123 established that no timeout value can bound it, and shipped the only two things that help — retrying past it, and serving from the ISR cache so the function is not invoked at all.

## What is not yet known

The cause of the stall is **not** identified. The obvious candidate is cold-instance JavaScript evaluation and GC at Lambda's memory-proportional CPU share, but that is inference. The facts in hand:

- `AWS_LAMBDA_FUNCTION_MEMORY_SIZE` is `1024`, so the function gets roughly 0.58 vCPU.
- V8 reports a heap limit of 1,018 MB, and RSS at rest is 675–795 MB. The instance is running close to its container.
- `process.env.NODE_OPTIONS` is `null` inside the function, so `netlify.toml`'s `--max-old-space-size=6144` is **not** leaking into the runtime and letting V8 overrun the container. That hypothesis is ruled out.

[...]

Issue #1117, opened 2026-08-09
PG_POOL_MAX=1 makes Promise.all over Prisma reads a no-op on every deployed context


teetangh:
## Summary

`PG_POOL_MAX=1` is set on production, deploy-preview and branch-deploy. `lib/prisma.ts` passes it straight through to `pg.Pool` as `max`, so every Netlify function instance holds exactly one Postgres client. Two Prisma queries issued concurrently therefore serialise on that single connection, and `Promise.all` over Prisma reads cannot save a round trip on any deployed context.

This matters because "these two reads are independent, so `Promise.all` them" is the standard fix reached for whenever a request shows a chain of dependent database waves, and on this deployment it is inert. It is worth writing down before someone spends another PR on it.

## How it was measured

`customSession` in `lib/auth.ts` ran its SSO enforcement check after `membership.findMany`, sequential purely by code order. The check needs only `user.email` and `user.id`, reads a disjoint set of tables (`OrgDomainClaim`, `OrganizationSSOSettings`, `SsoProvider`, `Account`) and writes nothing, so it was hoisted to the top of the callback and joined with `membership.findMany` in one `Promise.all`. That should have removed one cross-region wave from every authenticated request.

Deploy preview against the `dev` branch deploy, twenty order-balanced interleaved rounds against `GET /api/auth/get-session` with a real session for `zara.brown@yahoo.com`, both functions warmed first:

| | median | mean | min | p90 |
|---|---|---|---|---|
| `dev` (sequential) | 1,142 ms | 1,149 ms | 1,086 ms | 1,241 ms |
| preview (`Promise.all`) | 1,176 ms | 1,181 ms | 1,112 ms | 1,246 ms |

The parallel version is marginally slower, which is to say the difference is noise. Anonymous `get-session` measured 557 ms on `dev` and 558 ms on the preview, so the two deployments are directly comparable and there is no deployment-level offset hiding a win.

The change was reverted inside PR #1116 rather than shipped on theory.

## What to decide

The pool size is not obviously wrong. The comment in `lib/prisma.ts` explains it: `pg.Pool` defaults to ten clients per function instance, and at Netlify's 125 concurrent invocations that would dwarf Supavisor's client cap. `PG_POOL_MAX=1` is a deliberate trade of per-request latency for not exhausting the pooler.

Three questions worth answering with measurements rather than argument:

[...]

PR #1126, opened 2026-08-09
fix: restore production server diagnostics and stop paying for guaranteed-empty reads

### LINKED ISSUE #1121: Consultant home issues 9 guaranteed-empty user queries (WHERE id IN (NULL)) when a consultant has no upcoming slots
## Symptom

Sentry issue `FAMILIARISE_WEB-G` ("N+1 Query", prod, release 2737267b, `GET /dashboard/consultant/[consultantId]/home`) reports `prisma:client:operation` totalling **4,624 ms**, dominated by nine executions of:

```sql
SELECT ... FROM "public"."users" WHERE "users"."id" IN (NULL) OFFSET $1
```

`IN (NULL)` is what Prisma emits for an **empty** `in` list, so all nine queries are guaranteed to match zero rows.

> Sample size is 1 event and the Sentry issue itself was **not** independently re-read (see "Verification status"). The code path below, however, was verified against current `dev`.

## Cause — verified in code

`lib/data/consultant-dashboard.ts:358-370` builds the id list:

```ts
const homeAppointmentIds = [
  ...new Set(soonestSlots.map((s) => s.appointmentId)),
].slice(0, HOME_APPOINTMENTS_TAKE);
```

`soonestSlots` selects slots with `endsAt >= startOfToday`, so `homeAppointmentIds` is `[]` for any consultant with no current or upcoming slot — a new consultant, an entirely past book, or one whose upcoming work was cancelled or soft-deleted.

It is then passed unguarded at `lib/data/consultant-dashboard.ts:391-394`:

```ts
prisma.appointment.findMany({
  where: { id: { in: homeAppointmentIds } },
  include: appointmentInclude,
}),
```

There is no `homeAppointmentIds.length === 0` short-circuit anywhere in the file.

`appointmentInclude` (`lib/data/consultant-dashboard.ts:107-229`) contains exactly **nine** nested `user:` selections — at lines 111, 125, 134, 148, 157, 173, 183, 204 and 214. Prisma resolves each as its own follow-up `SELECT ... FROM users`, which is why the count is nine and not one. That matches the Sentry span count one-for-one.

## Impact

Not a correctness bug — an empty appointment list correctly renders nothing. It is pure waste: ~4.6 s of database time on a dashboard route to fetch a guaranteed-empty result, and it lands hardest on **brand-new consultants**, whose very first dashboard view is the slowest one.

This also compounds with `PG_POOL_MAX=1` (#1117): the nine statements serialise on a single connection, so statement count converts directly into wall time.

## Fix

Short-circuit when the id list is empty, and skip the dependent reads rather than issuing them. Kept out of PR #1118 deliberately — that PR is measurement-only.

## Verification status

- VERIFIED: the call site, the empty-array path, and the nine `user` includes, all against current `dev`.
- NOT VERIFIED: the Sentry issue itself. The Sentry MCP server requires an interactive OAuth flow and `SENTRY_AUTH_TOKEN` is secret-scoped on Netlify and empty locally, so the event could not be re-read first-hand. Timings above are transcribed from the issue report, sample size 1.

Also worth checking while here: `HomePageClient.tsx:17-22` sets `staleTime: 0`, overriding `STALE_TIMES.MEDIUM`, so the client immediately refetches `GET /api/dashboard/consultant/[id]` after hydration and recomputes the identical payload — potentially doubling the nine to eighteen per page view.

### LINKED ISSUE #1122: compiler.removeConsole strips server-side console.* too — ~993 production diagnostics are silently inert
## What happens

`next.config.mjs:232-235`:

```js
// Reduce JavaScript bundle size by removing console statements in production
compiler: {
  removeConsole: process.env.NODE_ENV === "production",
},
```

The comment frames this as a **bundle-size** measure, which reads as client-only. It is not — SWC strips `console.*` from **server** code as well. Every server-side diagnostic in this repo is therefore silently absent from production function logs.

## Verified empirically

A `console.warn` was added at Prisma client construction (`makeClient()`, runs once per instance boot) and deployed to preview 1118 (commit `7467228b`, deploy `6a787953f9b23a0008307b5f`). Eight concurrent requests were fired to force new instances.

- `[Prisma:INIT]` occurrences in the function log: **0**
- Module loads confirmed in the same window: **2**, via the Better Auth `Social provider github/facebook is missing clientId or clientSecret` pairs at `13:09:22.170` and `13:09:35.274`
- Third-party lines in the same window that *did* survive: those Better Auth warnings, and `prisma:error timeout exceeded when trying to connect`

Third-party output survives because `node_modules` is not compiled by SWC. Our own code is.

## Scope

**993** `console.*` call sites under `lib/` and `app/api` (excluding tests), all inert in production. These include the two files that `lib/prisma.ts:95-97` cites as establishing "the lib/ convention":

```
lib/maintenance-cron.ts:60,75,83,91,97
lib/redis.ts:34
```

Also affected, among others: `lib/stream-logger.ts`, `lib/meeting.ts`, `lib/email.ts`, `lib/auth.ts`, `lib/payments/index.ts`, `lib/trials/cancellation.ts`, `lib/supabase.ts`, `lib/betterstack.ts`.

## Why it matters

This is a silent-failure generator. A diagnostic added this way passes `tsc`, ESLint, Jest and local `next dev`, and then produces nothing in production — so it is worse than no diagnostic, because it looks like evidence of absence. It cost real time during #14: the instrumentation written for that PR had to be measured, found inert, and removed.

## Options

- Narrow it to `{ exclude: ["error", "warn"] }` so deliberate diagnostics survive while chatty `console.log` is still stripped.
- Or apply it to the client compilation only, since bundle size is the stated goal and server code does not ship to the browser.
- Or standardise on `Sentry.logger` for anything that must be observable — noting it ships to Sentry, not to the Netlify function log, so it does not serve someone reading `netlify logs`.

At minimum, fix the misleading comment and the "lib/ convention" claim in `lib/prisma.ts`.

Found while decomposing the cold TTFB — see PR #1118.

teetangh:
Three issues that share one root theme: the codebase could not see itself in production.

## 1. `removeConsole` was deleting every server diagnostic (#1122)

`next.config.mjs` set `removeConsole: process.env.NODE_ENV === "production"` under a comment framing it as a bundle-size measure, which reads as client-only. It is not. SWC applies the transform to the **server** layer as well, and Next offers no client-only scoping — [vercel/next.js#48410](https://github.com/vercel/next.js/discussions/48410) asks for one and is unresolved with no maintainer reply. So ~1,110 server-side `console.*` sites across 326 files were silently absent from the deployed function.

Proven rather than inferred, in #1122: a `console.warn` at Prisma client construction produced **0** log lines on preview 1118 across two independently confirmed module loads, while third-party output in the same window survived — `node_modules` is not compiled by SWC, our code is. This has cost real time twice; the #1124 event-loop investigation had to return its findings through the HTTP response body because nothing it logged reached the function log.

Narrowed to `{ exclude: ["error", "warn"] }`, the only documented knob. `console.log` stays stripped, which is the whole of the bundle saving. Two things worth recording:

- **Sentry cannot substitute.** `consoleLoggingIntegration()` patches `globalThis.console` at runtime; `removeConsole` deletes the call expressions at compile time. Adding it today would have recovered zero diagnostics.
- **The option will actually take effect here.** The two documented ways it silently no-ops are a Babel config and Turbopack. This repo has neither — no `.babelrc`, and `build` is plain `next build`.

`[Prisma:INIT]` now logs the resolved connect/query/pool budgets once per instance boot. That is the line #1118 wanted and never shipped, it makes deployed values readable from `netlify logs` instead of argued about, and it doubles as this change's A/B probe.

Three stale comments corrected: `lib/prisma.ts` cited a "lib/ convention" of `console.warn` while calling `Sentry.logger.warn` beneath it, and the two in `lib/data/fail-open.ts` — the only place in the repo that had noticed this behaviour at all.

## 2. Nine guaranteed-empty queries on consultant Home (#1121)

`homeAppointmentIds` is `[]` for any consultant with no current or upcoming slot — a new consultant, an entirely past book, or one whose upcoming work was cancelled. Prisma renders an empty `in` as `IN (NULL)` and still round-trips, and `appointmentInclude` carries exactly nine nested `user:` selections, each resolved as its own follow-up `SELECT`.

The Sentry issue (`FAMILIARISE_WEB-G`) was flagged in #1121 as never independently re-read; it has now been read first-hand. It holds — production, `GET /dashboard/consultant/[consultantId]/home`, `prisma:client:operation` at **4,624 ms**, that exact query nine times. One correction, though, from the span tree:

```
prisma:client:operation [4624ms]
  ├─ SELECT ... IN (NULL)   974ms      ← first query behind a connect
  ├─ pg-pool.connect        910ms
  ├─ SELECT ... IN (NULL)    63ms      ← same query, warm
  ├─ SELECT ... IN (NULL)   916ms
  ├─ pg-pool.connect        845ms
  └─ ...
```

The empty selects cost **63–68 ms** warm. The bulk is `pg-pool.connect` and the first query queued behind each one — the `PG_POOL_MAX=1` serialisation of #1117. So **statement count is the lever**, and not issuing the statements is the cheapest way to pull it. Caveat kept honest: 1 occurrence, 0 users impacted, first seen 2026-07-11 on release `2737267b`. The code path is verified against current `dev` and is now pinned by a unit test.

Guarded **per-query, not by early return** — the other 13 reads in that `Promise.all` are consultant-scoped rather than id-scoped and must still run, or Financial Summary goes blank. Three sibling unguarded `in:` sites get the same treatment (`explore-programs.ts` ×2, `explore-organisations.ts`), which matters more now that #1110 made those routes ISR. Uses the repo's existing inline-ternary convention rather than introducing a helper for four sites.

Also removes `HomePageClient`'s `staleTime: 0` override. It sat under a comment about showing stale data immediately, which is not what `staleTime` does: it marked the #890 SSR-prefetched entry stale on mount, so the client refetched `GET /api/dashboard/consultant/[id]` right after hydration and recomputed the identical payload — turning nine empty queries into eighteen. The factory's 2-minute value now applies, as it does for every sibling query on the dashboard.

## 3. The #1119 guard's blind spot (part of #1125)

`__tests__/explore/isr-routes-never-fail-open.test.ts` pins that degrading and being cacheable are mutually exclusive, but its detector is literally `/perRequest/.test(src)` — it can only see degradation that goes through the helper. A hand-written `catch { return [] }` is the same hazard and is invisible to it.

- The two swallows in `/explore/programs` now route through `fallbackOnTransientDbError` with `perRequest: true` (provable: that file exports `dynamic = "force-dynamic"`). They previously swallowed *every* error class, directly contradicting the contract the helper documents — a mapper or schema regression would have rendered an empty programs page indefinitely with no signal. This gives the helper its first call sites; it was dead code kept for symmetry.
- The local `try/catch` wrappers are gone rather than patched, so all six reads on that page now read identically. `fetchProgramStats` disappeared because it had become a pure passthrough.
- `app/api/announcements/route.ts` gets the `Cache-Control: no-store` its sibling `app/api/user/consultants/route.ts` documents. Not load-bearing today — the success path sets no cache header and Next 15 leaves Route Handlers uncached — but the two disagreed and this is the safe half.
- The guard gains a fourth assertion: **a cacheable route must contain no bare `catch {`**. Sharper than trying to regex "returns a placeholder" — a catch that binds nothing *cannot* report what it swallowed. Zero violations on current `dev`.

## Verification

Green CI is not verification here, so:

**Both new guards are mutation-checked.** Reverting `removeConsole` to the bare `NODE_ENV` check fails 2 of 4 assertions; adding a `catch {` to `app/page.tsx` (cacheable) fails the ISR guard; removing the empty-`in` guard fails the read-shape test. All three pass again on restore.

The bare-catch detector is anchored on **inline fixture strings**, not on real files — the #1125 sweep will delete the sites that would otherwise anchor it, and it would have gone vacuous silently.

`tsc --noEmit` clean, `eslint` clean on all changed files (warnings are blocking here), 299 tests green across `__tests__/explore`, `__tests__/dashboards`, `__tests__/lib`, `__tests__/security`. No local `next build` (RAM).

**Still to do on the preview:** warm the function, fire ~8 concurrent requests to force new instances, and grep the function log for `[Prisma:INIT]`. The identical probe measured 0 occurrences on preview 1118, so any non-zero count is the proof.

## Follow-up filed

Excluding `error` and `warn` makes ~709 server-side call sites start writing to Netlify function logs. Sentry is deliberately `sendDefaultPii: false`, but nothing constrains what a `console.error` passes, and several log whole error objects or request bodies. Tracked separately rather than blocking the return of production diagnostics.

Closes #1122
Closes #1121
Part of #1125

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dashboard and Explore performance by skipping unnecessary data queries when no results are available.
  - Improved handling of temporary database issues with safer, non-cacheable responses and graceful fallbacks.
  - Preserved correct filtering and ordering when ranked results are available.
  - Consultant dashboards now use consistent data freshness behavior.

- **Monitoring**
  - Production logs retain errors and warnings while removing routine log output.
  - Added clearer database connection and performance diagnostics.

### DIFF
--- a/__tests__/dashboards/consultant-home-read-shape.test.ts
+++ b/__tests__/dashboards/consultant-home-read-shape.test.ts
@@ -105,6 +105,34 @@ describe("consultant Home read shape (#1101)", () => {
     }
   });
 
+  it("issues no display read at all when there is nothing upcoming (#1121)", async () => {
+    // slotFindMany resolves [] from beforeEach, so homeAppointmentIds is empty —
+    // a new consultant, an entirely past book, or one whose upcoming work was
+    // cancelled. Prisma renders an empty `in` as `IN (NULL)`, and the display
+    // read's include graph then costs nine follow-up SELECTs on `users` for a
+    // result that is guaranteed to be empty.
+    await getConsultantDashboard("cp-1");
+
+    const displayReads = apptFindMany.mock.calls.filter((c) => c[0].include);
+    expect(displayReads).toEqual([]);
+
+    // The active-book read is id-independent and MUST still run — the guard is
+    // per-query, not an early return, or Financial Summary goes blank.
+    expect(apptFindMany.mock.calls.some((c) => c[0].select)).toBe(true);
+  });
+
+  it("still issues the display read when there IS something upcoming (#1121)", async () => {
+    // Non-vacuity anchor for the assertion above: prove the guard is keyed on
+    // emptiness and has not simply deleted the read.
+    slotFindMany.mockResolvedValue([{ appointmentId: "appt-1" }]);
+
+    await getConsultantDashboard("cp-1");
+
+    const displayReads = apptFindMany.mock.calls.filter((c) => c[0].include);
+    expect(displayReads).toHaveLength(1);
+    expect(displayReads[0][0].where.id.in).toEqual(["appt-1"]);
+  });
+
   it("derives active clients from a dedicated query, not the truncated display list", async () => {
     // Display list is capped; the active book is not. Deriving counts from the
     // capped array under-reported Financial Summary for any real consultant.
--- a/__tests__/explore/isr-routes-never-fail-open.test.ts
+++ b/__tests__/explore/isr-routes-never-fail-open.test.ts
@@ -41,6 +41,19 @@ const isForceDynamic = (src: string) =>
   /export\s+const\s+dynamic\s*=\s*["']force-dynamic["']/.test(src);
 const degrades = (src: string) => /perRequest/.test(src);
 
+// `degrades` above only sees degradation that goes through lib/data/fail-open.ts,
+// because `perRequest` is the helper's opt-in flag. A hand-written
+// `catch { return [] }` is the same hazard on a cacheable route and contains no
+// such token — that blind spot is #1125, and two live examples were found in
+// /explore/programs while reviewing #1123.
+//
+// The invariant pinned instead is narrower and sharper than "returns a
+// placeholder", which is not reliably detectable from source: a catch that binds
+// NOTHING cannot report what it swallowed, to Sentry or anywhere else. Binding
+// the error is the floor. `catch (e) {` passes; `catch {` does not. `.catch(fn)`
+// is untouched — that is a call, not a clause.
+const hasBareCatch = (src: string) => /(?:^|[^.\w])catch\s*\{/.test(src);
+
 describe("#1119 — a cacheable route must never fail open", () => {
   it("finds the route files it is supposed to be guarding", () => {
     // Guards against the walk silently matching nothing and passing vacuously.
@@ -64,6 +77,29 @@ describe("#1119 — a cacheable route must never fail open", () => {
           !isForceDynamic(f.src) &&
           !f.rel.endsWith("route.ts"),
       )
+      .map((f) => f.rel);
+    expect(offenders).toEqual([]);
+  });
+
+  it("the bare-catch detector matches the shape it claims to (#1125)", () => {
+    // Anchored on fixtures rather than on real files on purpose. The obvious
+    // anchor — "some route in the repo has a bare catch" — decays into a vacuous
+    // pass the moment the sweep in #1125 removes the last one, and it would do
+    // so silently, which is the exact failure mode this whole file exists for.
+    expect(hasBareCatch("try { x() } catch { return [] }")).toBe(true);
+    expect(hasBareCatch("try { x() } catch{return null}")).toBe(true);
+    expect(hasBareCatch("try { x() } catch (err) { report(err); return [] }")).toBe(
+      false,
+    );
+    // A `.catch(handler)` call is the sanctioned form and must not be flagged.
+    expect(hasBareCatch("read().catch(emptyOnTransientDbError('x'))")).toBe(
+      false,
+    );
+  });
+
+  it("no cacheable route swallows an error it cannot report (#1125)", () => {
+    const offenders = files
+      .filter((f) => isCacheable(f.src) && hasBareCatch(f.src))
       .map((f) => f.rel);
     expect(offenders).toEqual([]);
   });
--- a/__tests__/explore/isr-routes-never-fail-open.test.ts
+++ b/__tests__/explore/isr-routes-never-fail-open.test.ts
@@ -52,7 +52,13 @@ const degrades = (src: string) => /perRequest/.test(src);
 // NOTHING cannot report what it swallowed, to Sentry or anywhere else. Binding
 // the error is the floor. `catch (e) {` passes; `catch {` does not. `.catch(fn)`
 // is untouched — that is a call, not a clause.
-const hasBareCatch = (src: string) => /(?:^|[^.\w])catch\s*\{/.test(src);
+//
+// Comments are matched between `catch` and `{` because `\s` alone does not, and
+// `catch /* why */ {` is both legal and exactly what someone reaches for when
+// explaining why a swallow is fine. A guard a comment can switch off is worse
+// than no guard, because it still reads as coverage.
+const hasBareCatch = (src: string) =>
+  /(?:^|[^.\w])catch\s*(?:\/\/[^\n]*\n|\/\*[\s\S]*?\*\/|\s)*\{/.test(src);
 
 describe("#1119 — a cacheable route must never fail open", () => {
   it("finds the route files it is supposed to be guarding", () => {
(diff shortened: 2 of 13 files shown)


teetangh:
## Verified on the preview — `[Prisma:INIT]` now reaches the function log

The measurement this PR was waiting on. Same technique as #1122's original experiment, opposite result.

**Query** (identical for both, `netlify logs --source functions --function ___netlify-server-handler --since 25m --url <deploy>`):

| Deploy | `[Prisma:INIT]` occurrences |
|---|---|
| deploy-preview-1126 (`6a78d94e`, has the fix) | **2** |
| production `familiarisenow.com` (no fix) | **0** |

```
03:36:28.519 INFO  Duration: 34620 ms	Memory Usage: 962 MB
03:36:34.177 WARN  [Prisma:INIT] connect=3000ms query=6000ms poolMax=1 slowQuery=500ms buildPhase=false
03:36:34.218 WARN  [Better Auth]: Social provider facebook is missing clientId or clientSecret
03:36:34.218 WARN  [Better Auth]: Social provider github is missing clientId or clientSecret
03:36:48.639 WARN  [Prisma:INIT] connect=3000ms query=6000ms poolMax=1 slowQuery=500ms buildPhase=false
03:36:48.797 WARN  [Better Auth]: Social provider github is missing clientId or clientSecret
03:36:48.798 WARN  [Better Auth]: Social provider facebook is missing clientId or clientSecret
```

The Better Auth lines are the control, and they are why this is conclusive rather than suggestive. They come from `node_modules`, which SWC never compiled, so they survived even while our own diagnostics did not — #1122 used exactly them to prove **2 module loads occurred and produced 0 `[Prisma:INIT]`**. Here the same pairing appears, 41ms apart, **2 module loads → 2 `[Prisma:INIT]`**.

### Three things the line settles that were previously inferred

- `connect=3000ms` — the 3s **runtime** budget is the one selected, not the 30s build-phase branch.
- `buildPhase=false` — `NEXT_PHASE` is unset at runtime. #1120 concluded this indirectly (from the fact that a degraded page rendered at all); it is now directly observed.
- `poolMax=1` — #1117's `PG_POOL_MAX=1` confirmed live on deploy-preview, not just asserted from env config.

### Also caught, unrelated to this PR

`Duration: 34620 ms` on a cold instance, and a `/explore/experts` request that took 37.8s and returned 500. That is #1124 reproducing live, and the 500 is #1123 working as designed — rethrowing rather than caching a degraded render. Not a regression from this PR; the preview simply made it visible.

Client-side half checked separately: `console.log(` appears **0** times across the 26 client chunks the landing page loads, so the bundle-size goal the original option was written for is preserved.

Issue #1128, opened 2026-08-09
Invoice-refund wallet credit is called "guaranteed bookkeeping" but has no durability guarantee


teetangh:
## What was found

`app/api/webhooks/utils.ts`, inside `handleRefundCreated`'s invoice-refund branch, contains this note above the credit-note mint:

> For now, the compensating `WalletEntry` credit below plus the `INVOICE_REFUNDED` audit log is **the guaranteed bookkeeping**; the operator runbook calls out bookings that may need manual reversal.

Twenty lines later, that credit is wrapped in a `try`/`catch` whose only action was a `console.warn` — which `compiler.removeConsole` deleted from production entirely (#1122). So the one thing the function calls guaranteed was, in practice, best-effort and unobservable.

The catch's own comment said it was there to "swallow if no wallet flow applies", but the enclosing `if (ba && ba.fundingSource === "WALLET")` already handles that case structurally. The catch therefore only ever fired on a genuine failure.

#1126's follow-up adds `reportSentryError` there, so a failure is now visible. This issue is about the durability question that telemetry does not answer.

## Why it was not simply rethrown

Rethrowing looks like the obvious fix and is worse. The dispatcher stamps `error=true` on a throw, and the stuck-event sweeper only re-drives events with `error=null` (documented on `DeferSignal` at the top of the same file). So a rethrow would:

1. roll back the whole transaction — the invoice would not be marked `REFUNDED` and the GST credit note would not be minted, and
2. retire the webhook event permanently, since nothing re-drives an `error=true` row.

A booked refund missing its wallet credit is recoverable by hand. A refund that never books and never retries is not.

## What to decide

1. **What failure modes can actually reach that catch.** Worth being precise rather than assuming: a Postgres-level error inside the interactive transaction puts the transaction into an aborted state, so the swallow does not let it commit — it resurfaces at `COMMIT`. A JavaScript-level failure inside `walletCredit` before any SQL is issued *does* let the transaction commit without the credit. Only the second is silently lossy, and it should be confirmed rather than reasoned about.
2. **Whether `DeferSignal` is the right vehicle.** It exists precisely for "processable, but not completable yet" and gets re-driven by the sweeper. A wallet-credit failure is arguably that shape, but deferring the whole refund to retry a compensating credit needs the idempotency of every preceding step checked first — the invoice status short-circuit and `mintInvoiceRefundCreditNote` are both idempotent on `refundId`, which is promising.
[...]

Issue #1127, opened 2026-08-09
Audit what console.error/warn payloads serialise, now that they reach Netlify function logs


teetangh:
## Context

#1122 (fixed in #1126) narrowed `compiler.removeConsole` to `{ exclude: ["error", "warn"] }`. Before that, SWC deleted every server-side `console.*` from the deployed function, so ~1,110 diagnostics produced nothing in production.

That was the right thing to fix — zero production diagnostics had already cost two investigations. But it has a consequence that was accepted deliberately rather than overlooked, and is recorded here so it does not get rediscovered as a surprise.

## The concern

Roughly **709** server-side call sites under `app/api/**` and `lib/**` are at `error` or `warn` level and now write to Netlify function logs:

| Level | `app/api` + `lib` + `actions` |
|---|---|
| `console.error` | 571 |
| `console.warn` | 138 |

Nothing constrains what those calls pass. Several serialise whole error objects or request bodies, and a Prisma error in particular can carry the parameters of the failing query. This sits against a Sentry config that is deliberately `sendDefaultPii: false` (`sentry.shared.config.ts:38`) — so the two observability channels now have materially different PII postures, which was not a considered decision.

Netlify function logs are team-scoped rather than public, which is why this was judged an acceptable trade against the alternative of no diagnostics at all. It is not zero risk.

## What to do

1. Sweep `console.error` / `console.warn` in `app/api/**`, `lib/**` and `actions/**` for arguments that can carry personal data — request bodies, whole `error` objects from Prisma, session or user objects, email addresses, phone numbers, payment identifiers.
2. Decide a house rule for what a server-side error log may contain, and write it down next to the `removeConsole` config so the next person does not have to infer it.
3. For the paths that need structured detail, prefer `reportSentryError` (`lib/observability/report.ts`) or `Sentry.logger`, which already have a scrubbing posture, over a raw `console.error`.
4. Consider whether `lib/logging/webhook-scrub.ts` generalises — it already exists for exactly this problem on the webhook path.

## Not urgent

No known leak, no incident. This is a posture question opened by a deliberate trade, filed so the trade stays visible.

Follows #1122. Related: #900 (Sentry observability), #1125 (silent catch sweep).

PR #1129, opened 2026-08-09
fix: report the silent swallows that lose money, access and availability

### LINKED ISSUE #1125: Bare catch-and-placeholder blocks are the same hazard as fail-open, and the new #1119 guard cannot see them

teetangh:
> **Stacked on #1126.** Base is `fix/server-diagnostics-and-wasted-reads`, not `dev` — retarget after #1126 merges. It is stacked because its triage rule is only correct once #1126 lands: `lib/supabase.ts:215` and four siblings do `catch (e) { console.error(...); return false }`, which *looks* reported in source and is byte-identical to a bare catch in production until `removeConsole` is narrowed.

## How this was scoped

#1125 asked for a sweep. A six-batch read-only census over disjoint directories found **229** catch-and-placeholder sites, 29 of them rated HIGH.

Rather than apply that, **19 of the 29 HIGH calls were checked against source first. Roughly a third survived.** Two claimed a swallow was unreported in code that already calls `Sentry.captureException`. Several described degradations that are correct *and* already return a failure to their caller — `lib/email.ts:421` dead-letters and a worker replays it; `lib/novu/service.ts:177/:211` both return `{success:false}`; `drain-sessions.ts:150` surfaces through `result.errors` to the operator.

So only the survivors are fixed here. The full reviewed inventory is posted on #1125 with an explicit caveat about its reliability, so the tail stays actionable without becoming an unreviewable diff.

**A theme runs through the ones that were real.** Each carries a comment describing a case its own code already handles structurally — which is what made the catch look reasonable while it fired *exclusively* on genuine faults.

### Money — invoice-refund wallet credit (`app/api/webhooks/utils.ts`)

Sat under *"Opportunistic … Swallow if no wallet flow applies"*. But `if (ba && ba.fundingSource === "WALLET")` already gates that, so the catch only ever fired on a real failure — and wrote a `console.warn` that production deleted. Twenty lines above, the code calls this credit **"the guaranteed bookkeeping"** for a refunded invoice.

**Deliberately still not rethrown.** The dispatcher stamps `error=true` on a throw and the stuck-event sweeper only re-drives `error=null` (documented on `DeferSignal` at the top of the same file), so rethrowing would roll back the whole refund booking *and* retire the event permanently. A booked refund missing its wallet credit is recoverable by hand; a refund that never books and never retries is not. Reported loudly instead. Durability is **#1128**.

### Security — SSO enforcement failed open (`lib/auth.ts`)

`catch { /* non-fatal — don't break session */ }` leaves `ssoEnforcementFailed` false, so the session **passes** enforcement. The two fail-opens reasoned about in the comment directly above it are deliberate; this one never was, and was silent.

### Security — rate limiting (`lib/rate-limit.ts`)

`catch { return null; // Redis down — fail open }`, and `null` means *not rate limited*. A Redis outage silently disabled every limiter in the app. Fail-open kept (deliberate); now reported.

Note `identifier` is **deliberately not attached**: callers key on whatever identifies the caller and `app/api/consultants/search/route.ts:10` passes a raw client IP, which would put PII in Sentry against this project's `sendDefaultPii: false` (#1127). It buys nothing anyway — when Redis is down, every limiter fails, and the captured transaction already names the route.

### Truthfulness — referral landing (`app/r/[code]/page.tsx`)

Swallowed `applyReferralCode` under *"if code application fails (already referred, self-referral, etc.)"*. That function **returns `null`** for every one of those cases (self-referral, already referred, cap reached, invalid code) and only **throws** on a genuine fault — a serializable retry giving up, the 10s transaction timeout, the pooler. So the catch never saw the cases it named. It then redirected to `?ref_applied=true` unconditionally, asserting a reward the user might not have got. Now conditional on the return value.

### Access control — collaborator revocation (`lib/collaborators/service.ts`)

`removeUserFromEventChannel` reports failure by **returning `{ success: false }`**, not by throwing. It was awaited without reading that, so a failed chat-access revocation was indistinguishable from a success and the outer catch never fired — a removed collaborator kept chat access on every event. Results are now checked per event; the coordination channel's `.catch(() => {})` reports separately.

Also `removeCollaborator`'s profile lookup: a `.catch(() => null)` there skips **both** the notification and the revocation.

### Data loss — availability save (`utils/schedule/formatting.ts`) — the one behaviour change

`formatSlotsForApi` feeds a **PUT body** (`SettingsTab.tsx:457`), so degrading does not mean render less, it means **save less**. It carried three nested catches: a per-slot one dropped the offending slot from the payload, and an outer one returned `[]` for the entire schedule — after which SettingsTab refetched and displayed the wiped availability as *"what was actually saved"*.

All three are gone. `formatCustomSlot` now throws rather than returning a `null` the caller filtered away — that null came from `convertTimezoneToUtcWithOvernight`'s `""` sentinel, so it was indistinguishable from "this slot is fine but omitted". An availability save is all-or-nothing; the caller already had a destructive toast to fail into, and now reports to Sentry too.

`isValid: false` rows keep being filtered — that is the editor saying a row is incomplete, which is an answer, not an error. A test pins the distinction.

### Also

`isDbHealthy` (`app/api/webhooks/utils.ts`) — bare catch. Fails closed so no events are lost, but a persistent non-connectivity fault (schema drift) 503s every webhook indefinitely, invisibly.

`app/api/trials/[trialId]/route.ts` PATCH + DELETE — both call a refund *after* the trial writes commit, and both returned 500 with `console.error` and no Sentry, unlike the `apiError()` house pattern. This is the #1074 shape: a cancelled-but-unrefunded trial with no alert.

## Verification

`tsc --noEmit` clean, `eslint` clean on all changed files, **full suite 232 suites / 2595 tests green**.

Five new tests pin `formatSlotsForApi`'s contract, **mutation-checked**: reinstating the outer `catch { return [] }` fails exactly the two throw assertions and nothing else.

Every behaviour-preserving site was diffed by hand to confirm control flow is unchanged — fail-open still fails open, `isDbHealthy` still returns false, both trials handlers still 500, the profile lookup still returns null.

Closes #1125

## Summary by CodeRabbit

* **Bug Fixes**
  * Schedule updates now fail clearly when slot dates cannot be processed, preventing partial or empty schedules from being saved.
  * Invalid slots marked by the user continue to be excluded as expected.
  * Referral links now redirect accurately based on whether the referral was successfully applied.

* **Reliability**
  * Improved error tracking for trial actions, refunds, settings saves, authentication, collaboration tools, rate limiting, and referral processing to support faster issue resolution.

### DIFF
--- a/__tests__/schedule/format-slots-for-api-throws.test.ts
+++ b/__tests__/schedule/format-slots-for-api-throws.test.ts
@@ -1,0 +1,85 @@
+/**
+ * @jest-environment node
+ */
+
+/**
+ * #1125 — `formatSlotsForApi` feeds a PUT body (SettingsTab.tsx:457), so
+ * "degrade" here does not mean render less, it means SAVE less. It used to carry
+ * three nested catches: a per-slot one dropped the offending slot from the
+ * payload, and an outer one returned `[]` for the entire schedule — after which
+ * SettingsTab refetched and displayed the wiped availability as "what was
+ * actually saved". These pin the replacement contract: all of it, or an error
+ * the caller can show the consultant.
+ */
+import { formatSlotsForApi } from "@/utils/schedule/formatting";
+import type { SlotsType } from "@/utils/schedule/types";
+
+const slot = (startTime: string, endTime: string) => ({
+  id: `${startTime}-${endTime}`,
+  startTime,
+  endTime,
+  isValid: true,
+});
+
+describe("formatSlotsForApi is all-or-nothing (#1125)", () => {
+  it("formats every weekly slot it is given", () => {
+    const slots: SlotsType = {
+      monday: [slot("09:00", "10:00"), slot("14:00", "15:00")],
+      tuesday: [slot("11:00", "12:00")],
+    };
+
+    const result = formatSlotsForApi(slots, true, "UTC");
+
+    expect(result).toHaveLength(3);
+  });
+
+  it("formats every custom slot it is given", () => {
+    const slots: SlotsType = {
+      "2026-09-01": [slot("09:00", "10:00")],
+      "2026-09-02": [slot("13:00", "14:30")],
+    };
+
+    const result = formatSlotsForApi(slots, false, "UTC");
+
+    expect(result).toHaveLength(2);
+  });
+
+  it("throws rather than dropping a custom slot whose date key is unusable", () => {
+    // The old behaviour returned the OTHER day's slots and silently omitted
+    // this one, so the consultant saved a schedule missing a day they had set.
+    const slots: SlotsType = {
+      "2026-09-01": [slot("09:00", "10:00")],
+      "not-a-date": [slot("09:00", "10:00")],
+    };
+
+    expect(() => formatSlotsForApi(slots, false, "UTC")).toThrow(
+      /Invalid date format/,
+    );
+  });
+
+  it("throws rather than wiping the whole schedule when a day key is unusable", () => {
+    // The outer catch turned this into `[]`, which is indistinguishable from
+    // "this consultant has no availability" once it reaches the API.
+    const slots: SlotsType = {
+      notaday: [slot("09:00", "10:00")],
+    };
+
+    expect(() => formatSlotsForApi(slots, true, "UTC")).toThrow();
+  });
+
+  it("still skips slots the caller already marked invalid", () => {
+    // Distinct from a formatting FAILURE: `isValid: false` is the editor saying
+    // the row is incomplete, which is an answer, not an error. Those must keep
+    // being filtered rather than throwing.
+    const slots: SlotsType = {
+      monday: [
+        slot("09:00", "10:00"),
+        { ...slot("bad", "worse"), isValid: false },
+      ],
+    };
+
+    const result = formatSlotsForApi(slots, true, "UTC");
+
+    expect(result).toHaveLength(1);
+  });
+});
--- a/__tests__/schedule/format-slots-for-api-throws.test.ts
+++ b/__tests__/schedule/format-slots-for-api-throws.test.ts
@@ -67,6 +67,31 @@ describe("formatSlotsForApi is all-or-nothing (#1125)", () => {
     expect(() => formatSlotsForApi(slots, true, "UTC")).toThrow();
   });
 
+  it("throws rather than dropping a WEEKLY slot whose UTC conversion fails", () => {
+    // The weekly twin of the custom-slot case, and the worse half: weekly is the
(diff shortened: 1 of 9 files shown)


teetangh:
@coderabbitai review

